PR #3, opened 2018-04-27
Fix Android audio stream demo

ferjm:


src/backends/gstreamer/audio_stream.rs
```diff
@@ -1,31 +1,54 @@
...
 }
 
 impl GStreamerAudioStream {
     pub fn new() -> Result<Self, ()> {
-        let player = gst_player::Player::new(None, None);
-        player
-            .set_property("uri", &glib::Value::from("webaudiosrc://foo"))
-            .expect("Cant't set URI property");
-        Ok(Self { player: player })
+        if let Some(category) = gst::DebugCategory::get("gst-player") {
```
philn: This is useless now, sorry I missed it in the previous patch :) Anyway just a nit

PR #6, opened 2018-05-17
First bits of audio graph

ferjm:
This moves the hard-coded oscillator out of the src element and the audio processing to its own thread.


ferjm:
r? @Manishearth 

We can chat about this tomorrow if you have any doubts about the patch. Thanks!

ferjm:
@philn I would also love to hear your feedback on this. Thanks :)

Manishearth:
overall the design of the sink may need to change but we should first do the graph before we try this.

PR #13, opened 2018-05-19
Media thread

ferjm:
At some point we may want to use a dedicated process instead, but for now a thread for all media related stuff seems enough and paves the way for the potential upgrade to a process.


Manishearth:
Do we need a mutex here? The way I see it this will be run off an event loop so there doesn't have to be a mutex.

PR #15, opened 2018-05-22
Play silence by default

ferjm:
Quoting the [spec](https://webaudio.github.io/web-audio-api/#AudioScheduledSourceNode):

> Before a source is started (by calling start(), the source node MUST output silence (0). After a source has been stopped (by calling stop()), the source MUST then output silence (0).

### DIFF
--- a/src/backends/gstreamer/audio_sink.rs
+++ b/src/backends/gstreamer/audio_sink.rs
@@ -6,7 +6,6 @@ use audio::sink::AudioSink;
 use gst;
 use gst::prelude::*;
 use std::sync::Arc;
-use byte_slice_cast::*;
 
 // XXX Define own error type.
 
@@ -63,9 +62,13 @@ impl AudioSink for GStreamerAudioSink {
                 buffer.set_duration(next_pts - pts);
                 
                 let mut chunks = graph_.process(rate);
+                // sometimes nothing reaches the output
+                if chunks.len() == 0 {
+                    chunks.blocks.push(Default::default());
+                    info.format_info().fill_silence(chunks.blocks[0].as_mut_byte_slice());
+                }
                 debug_assert!(chunks.len() == 1);
-                let data = &mut chunks.blocks[0].data;
-                let data = data.as_mut_byte_slice().expect("casting failed");
+                let data = chunks.blocks[0].as_mut_byte_slice();
 
                 // XXXManishearth if we have a safe way to convert
                 // from Box<[f32]> to Box<[u8]> (similarly for Vec)
--- a/src/audio/block.rs
+++ b/src/audio/block.rs
@@ -1,4 +1,5 @@
 use smallvec::SmallVec;
+use byte_slice_cast::*;
 
 // defined by spec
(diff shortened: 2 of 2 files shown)


src/backends/gstreamer/audio_sink.rs
```diff
@@ -63,15 +40,12 @@ impl AudioSink for GStreamerAudioSink {
...
         let mut sample_offset = 0;
-        let n_samples = settings.samples_per_buffer as u64;
-        let buf_size = (n_samples as usize) * (info.bpf() as usize);
         let rate = info.rate();
 
         let graph_ = graph.clone();
-        let need_data = move |app: &AppSrc, _bytes: u32| {
-            let mut buffer = gst::Buffer::with_size(buf_size).unwrap();
+        let need_data = move |app: &AppSrc, bytes: u32| {
+            let mut buffer = gst::Buffer::from_slice(&SILENCE[0..bytes as usize]).unwrap();
```
sdroege: Here you create a buffer with read-only memory ...

src/backends/gstreamer/audio_sink.rs
```diff
@@ -81,19 +55,16 @@ impl AudioSink for GStreamerAudioSink {
...
                     .unwrap()
                     .into();
-                let next_pts: gst::ClockTime = (sample_offset + n_samples)
+                let next_pts: gst::ClockTime = (sample_offset + WEBAUDIO_RENDER_QUANTUM)
                     .mul_div_floor(gst::SECOND_VAL, rate as u64)
                     .unwrap()
                     .into();
                 buffer.set_pts(pts);
                 buffer.set_duration(next_pts - pts);
                 let mut map = buffer.map_writable().unwrap();
```
sdroege: ... and here the read-only memory would be copied into writeable memory so you can modify it.

It would be more efficient to create a buffer with writeable memory to begin with. `with_size` as before would do that, you only have to fill it with silence then e.g. with [this](https://sdroege.github.io/rustdoc/gstreamer/gstreamer_audio/struct.AudioFormatInfo.html#method.fill_silence)

src/backends/gstreamer/audio_sink.rs
sdroege: Also you use the bytes as passed in by appsrc. In theory that might be more than 4096, in theory it might not be a multiple of the unit size (sample size * channels), and worst: the alignment of the silence `Vec` might not be correct (it needs to be aligned to the sample size). You probably want to allocate always the same buffer size here like before and not worry about anything :)

ferjm:
Thank you for the feedback @sdroege! I just updated the patch.

Manishearth:
Okay, so there are two separate issues here.

One is "what should happen if we don't connect a source node to the output". That should be fixed by fill_silence. I've made this commit do so.

The other is "source nodes should output silence when stopped". That should be fixed by having them directly output silence 😄 . We can do that once we have messaging.

ferjm:
> The other is "source nodes should output silence when stopped". That should be fixed by having them directly output silence 😄 . We can do that once we have messaging.

I'm doing this as part of #21

PR #17, opened 2018-05-22
Use block/chunk abstraction for processing

Manishearth:


Manishearth:
I replaced the hashmap with a vec because there's no guaranteed order for hashmap iteration.

When we add graph support we'll replace it with a petgraph graph.

src/backends/gstreamer/audio_sink.rs
```diff
@@ -87,9 +61,13 @@ impl AudioSink for GStreamerAudioSink {
...
                 buffer.set_duration(next_pts - pts);
-                let mut map = buffer.map_writable().unwrap();
-                let data = map.as_mut_slice();
-                graph_.process(data, rate);
+                
+                let mut chunks = graph_.process(rate);
+                debug_assert!(chunks.len() == 1);
+                let data = &mut chunks.blocks[0].data;
+                let data = data.as_mut_byte_slice().expect("casting failed");
+                buffer.copy_from_slice(0, data).expect("copying failed");
```
sdroege: This looks like a potentially unneeded copy. You can wrap a buffer around any `AsRef<[u8]>` or `AsMut<[u8]>` if that helps anything here

Manishearth: How would I do that?

sdroege: Currently there is only minimal support for custom memory types in the bindings (see [here](https://github.com/sdroege/gstreamer-rs/issues/12)), but it's probably sufficient for what you need. [From a mutable slice](https://github.com/sdroege/gstreamer-rs/blob/1192a027b40c5ad46cd998cfa7fd11bc527bd1a9/gstreamer/src/buffer.rs#L75-L95) and [from an immutable slice](https://github.com/sdroege/gstreamer-rs/blob/1192a027b40c5ad46cd998cfa7fd11bc527bd1a9/gstreamer/src/buffer.rs#L97-L117)

Manishearth: Ah. Due to the way rustdoc deals with typedefs those functions don't show up in the docs where you'd expect 😄

sdroege: Yeah, personally I'm not using rustdoc because of that for anything built on top of gtk-rs. Easier to navigate the sources :)

Manishearth:
Do you have a list of features that would make rustdoc pleasant here?

I can try and implement them on the side, or mentor others to do so. Not
urgent, but worth fixing anyway.

The typedef one has been a constant paper cut for me in multiple codebases.

On May 22, 2018 11:39 PM, "Sebastian Dröge" <notifications@github.com>
wrote:

*@sdroege* commented on this pull request.
------------------------------

In src/backends/gstreamer/audio_sink.rs
<https://github.com/servo/media/pull/17#discussion_r190136441>:

> @@ -87,9 +61,13 @@ impl AudioSink for GStreamerAudioSink {
                     .into();
                 buffer.set_pts(pts);
                 buffer.set_duration(next_pts - pts);
-                let mut map = buffer.map_writable().unwrap();
-                let data = map.as_mut_slice();
-                graph_.process(data, rate);
+
+                let mut chunks = graph_.process(rate);
+                debug_assert!(chunks.len() == 1);
+                let data = &mut chunks.blocks[0].data;
+                let data = data.as_mut_byte_slice().expect("casting failed");
+                buffer.copy_from_slice(0, data).expect("copying failed");

Yeah, personally I'm not using rustdoc because of that for anything built
on top of gtk-rs. Easier to navigate the sources :)

—
You are receiving this because you authored the thread.
<https://github.com/servo/media/pull/17#discussion_r190136441>, or mute the
thread

sdroege:
> Do you have a list of features that would make rustdoc pleasant here?

The typedef one is the most annoying one, I'll try remembering others. Guillaume Gomez probably remembers some more too :) It improved quite a bit over the last months though.

Manishearth:
e? @ferjm

src/audio/graph_thread.rs
```diff
@@ -1,24 +1,25 @@
...
+    CreateNode(AudioNodeType),
     ResumeProcessing,
     PauseProcessing,
 }
 
 pub struct AudioGraphThread {
     // XXX Test with a hash map for now. This should end up
     // being a graph, like https://docs.rs/petgraph/0.4.12/petgraph/.
-    nodes: RefCell<HashMap<usize, Box<AudioNodeEngine>>>,
+    nodes: RefCell<Vec<Box<AudioNodeEngine>>>,
```
ferjm: Sorry, I forgot to mention this. Can we also leave the HashMap until we use something like [petgraph.GraphMap](https://docs.rs/petgraph/0.4.12/petgraph/graphmap/struct.GraphMap.html), please? We need it to implement node methods like #21.

ferjm: I meant leaving both, the HashMap and the Vec. Or maybe use something like (linked_hash_map)[https://docs.rs/linked-hash-map/0.5.1/linked_hash_map/]. But I guess Vec indices would work for now.

src/backends/gstreamer/audio_sink.rs
Manishearth: Okay so it looks like I'd need to convert between `Box<[u8]>` and `Box<[f32]>` here (and in future versions, between vecs), and byte-slice-cast doesn't support that yet. Filing a bug at https://github.com/sdroege/byte-slice-cast/issues/1 . It's not hard to write but I don't want to do one-off unsafe code in servo-media for it.

PR #25, opened 2018-05-23
DestinationNode placeholder

### LINKED ISSUE #24: Enable the creation of engines for AudioDestinationNodes

ferjm:
Fixes #24


Manishearth:
kinda feel like it shouldn't have an output at all, but having it in the graph and ensuring all sinks are destination node would work too.

(Also it lets us upmix/downmix with the same mechanism)

PR #26, opened 2018-05-23
Set up very basic messaging

Manishearth:
SetFloatParam is a dummy message, but the idea is to have this framework in place so that if I start working on AudioParams it won't clash with SchedulerNode work.

r? @ferjm


examples/play.rs
```diff
@@ -13,7 +13,10 @@ fn main() {
         options.gain = 0.5;
         graph.create_node(AudioNodeType::GainNode(options));
         graph.resume_processing();
-        thread::sleep(time::Duration::from_millis(5000));
+        thread::sleep(time::Duration::from_millis(2000));
```
ferjm: :) at this point I think I can reproduce the sine wave with my own voice

examples/play.rs
```diff
@@ -13,7 +13,10 @@ fn main() {
...
+        graph.message_node(0, AudioNodeMessage::SetFloatParam(220.));
```
ferjm: How would the API look like for two-way messaging (e.g. a getter)?

Manishearth: We probably send it a channel to respond through.

PR #28, opened 2018-05-23
Have the graph thread drive processing, make need_data use message passing

### LINKED ISSUE #27: Thread safety and driving model
So currently we're not actually thread safe. There's an unsafe impl of Send/Sync on AudioGraphThread, which shouldn't exist.

We simultaneously need:

 - mutable access from the graph thread for things like adding nodes and play/pause
 - mutable access from gstreamer for running the graph.

This isn't good. We could solve it with locks, but that introduces two competing synchronization mechanisms. Instead I think we should have everything run by the graph thread (IIRC this was the original plan but we didn't really explore it), with gstreamer being isolated in a different thread requesting data.

The plan would be:

 - The graph thread is the only thing that touches the graph
 - when the sink needs data, it requests it from the graph thread, and waits
 - the graph thread processes incoming graph change messages, as well as incoming need_data messages. Whenever the source needs data, it sends it over.

We don't actually need a separate "gstreamer thread" setup, the graph thread can own the gstreamer sink and the need_data callbacks let this work asynchronously.

I'm a bit afraid that the render quantum of ~3ms may be affected by this model, though.

Manishearth:
fixes #27

r? @ferjm

cc @sdroege

### DIFF
--- a/src/audio/graph_thread.rs
+++ b/src/audio/graph_thread.rs
@@ -72,8 +72,10 @@ impl AudioGraphThread {
     }
 
     pub fn event_loop(&self, event_queue: Receiver<AudioGraphThreadMsg>) {
+        let mut queued = -1;
         loop {
             if let Ok(msg) = event_queue.try_recv() {
+                queued = 0;
                 match msg {
                     AudioGraphThreadMsg::CreateNode(node_type) => {
                         self.create_node(node_type);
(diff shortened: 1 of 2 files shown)


ferjm:
I am going to try to implement [this approach](https://github.com/servo/media/issues/27#issuecomment-391642600) so we can compare and choose.

Issue #21, opened 2018-05-23
Implement support for AudioScheduledSourceNode start and stop methods


ferjm:
- [AudioScheduledSourceNode.start](https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-start)
- [AudioScheduledSourceNode.stop](https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-stop)

ferjm:
This will likely land first without scheduling support, until #22 is done.

PR #29, opened 2018-05-24
Fix audio render thread thread safety issues

ferjm:
…d tick

This is a different approach for #27.

There is one issue here that can be noticeable running the `play` example. It is supposed to play 2 seconds of two sine waves with different frequencies. With this patch the first sine wave steals most of the time of the second sine wave. I believe it is caused by the difference between the JS clock and the (currently not existent) WebAudio clock (there is a [nice blog post about this](https://www.html5rocks.com/en/tutorials/audio/scheduling/)). If that's the case this should be fixed by using the scheduling feature of AudioScheduledSourceNodes in the play example once #21 is done.

### DIFF
--- a/src/audio/graph_thread.rs
+++ b/src/audio/graph_thread.rs
@@ -6,7 +6,6 @@ use audio::oscillator_node::OscillatorNode;
 use audio::sink::AudioSink;
 use std::cell::RefCell;
 use std::sync::mpsc::Receiver;
-use std::sync::Arc;
 
 #[cfg(feature = "gst")]
 use backends::gstreamer::audio_sink::GStreamerAudioSink;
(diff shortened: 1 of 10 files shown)


ferjm:
@Manishearth @sdroege WDYT of this approach?

src/backends/gstreamer/audio_sink.rs
```diff
@@ -100,6 +67,61 @@ impl AudioSink for GStreamerAudioSink {
...
         let _ = self.pipeline.set_state(gst::State::Paused);
     }
+
+    fn has_enough_data(&self) -> bool {
+        self.appsrc.get_current_level_bytes() >= self.appsrc.get_max_bytes()
+    }
+
+    fn push_data(&self, mut chunk: Chunk) -> Result<(), ()> {
+        let sample_rate = self.sample_rate.get() as u64;
+        let ref audio_info = self.audio_info.borrow().clone().unwrap();
```
Manishearth: We shouldn't need to clone here. Call `.as_ref().unwrap()`

src/audio/graph_thread.rs
```diff
@@ -91,6 +93,12 @@ impl AudioGraphThread {
                     }
                 }
             }
+
+            // Process a render quantum if the sink can handle
+            // more data.
+            if !self.sink.has_enough_data() {
```
Manishearth: There's a slight problem here; this loop will eat up CPU instead of blocking because there's nothing to block on.

ferjm: Hmm, isn't this what's supposed to happen for the rendering thread? In any case, we could use appsrc's block property here instead of the custom `has_enough_data` check.

ferjm:
The other option may be to use the previous `has_enough_data` and sleep the thread for a few ms if the buffer is full to avoid 100% CPU usage.

Manishearth:
This can work.

I'm not sure which approach is better at this point. I guess avoiding the
synchronization overhead is good. Do you/@sdroege have opinions?

On May 24, 2018 2:28 PM, "Fernando Jiménez Moreno" <notifications@github.com>
wrote:

The other option may be to use the previous has_enough_data and sleep the
thread for a few ms if the buffer is full to avoid 100% CPU usage.

—
You are receiving this because you commented.
<https://github.com/servo/media/pull/29#issuecomment-391868838>, or mute
the thread

ferjm:
Until we have performance tests or we can manually test with more nodes or real WebAudio apps, it's hard to make an informed decision here, but I think the `has_enough_data` + sleep(5ms), while it's not very elegant, it's a good middle ground solution. We have no messaging overhead, the render thread can continue processing control thread messages and the blocking situations are controlled and for a known and configurable amount time.

Manishearth:
Sounds good. We can continue with that model then :smile:

ferjm:
I want to try yet another approach, but I am not sure if it'll work.

I think we can have an extra thread where we listen for appsrc callbacks. We can share the control <-> render thread channel with this thread. If the appsrc buffer is full, we will block on the control <-> render thread channel reception. When an appsrc.enough_data callback is triggered, a message will be sent through this channel to unblock and switch to the previous non-blocking event loop.

sdroege:
> but I think the has_enough_data + sleep(5ms)

I didn't review anything in detail here yet (later!) but a) sleeping 5ms will drift slowly if done just like that as it does not consider the amount of time taken for the actual code, b) you'll sleep according to the system clock and not according to the audio clock, so you either produce too much or too little data over a longer time (the audio clock will run at different speed).

ferjm:
I just pushed the commit with the approach I mentioned above. We no longer sleep the thread and we switch between a non-blocking and a blocking event loop depending on the audio sink data needs

PR #31, opened 2018-05-28
AudioScheduledSourceNode

### LINKED ISSUE #21: Implement support for AudioScheduledSourceNode start and stop methods

ferjm:
Fixes  #21.

### DIFF
--- a/src/audio/gain_node.rs
+++ b/src/audio/gain_node.rs
@@ -1,9 +1,12 @@
-use audio::node::AudioNodeMessage;
-use audio::param::Param;
-use audio::block::Tick;
-use audio::node::BlockInfo;
-use audio::node::AudioNodeEngine;
 use audio::block::Chunk;
+use audio::block::Tick;
+use audio::node::AudioNodeEngine;
+use audio::node::BlockInfo;
+use audio::param::{Param, UserAutomationEvent};
+
+pub enum GainNodeMessage {
+    SetGain(UserAutomationEvent),
+}
 
 pub struct GainNodeOptions {
     pub gain: f32,
@@ -11,9 +14,7 @@ pub struct GainNodeOptions {
 
 impl Default for GainNodeOptions {
     fn default() -> Self {
-        GainNodeOptions {
-            gain: 1.
-        }
+        GainNodeOptions { gain: 1. }
     }
 }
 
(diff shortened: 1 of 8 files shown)


Manishearth:
Overall looks good.




Also, is it possible to schedule multiple events? If so we'd need to store these as a list of events (in a smallvec probably) and check them each time.

ferjm:
> Also, is it possible to schedule multiple events? If so we'd need to store these as a list of events (in a smallvec probably) and check them each time

Good point! Yes, I believe it's possible.

Manishearth:
Also, we should ensure that these events only need to be checked per-block
(not frame) when we handle the timings.

Otherwise we need to make this part of an updateparameters step.

AudioParams need something like this too, so if you want we can leave it as
is as a k-type schedule till I add that support.

Manishearth:
Given that we now have ticks, can we internally use ticks for this? We can convert time floats to ticks at the user API boundary

ferjm:
@Manishearth I think this is ready for a final review. I checked the spec and found that we cannot schedule more than one start or stop times, so my believes [here](https://github.com/servo/media/pull/31#issuecomment-392723901) where incorrect.

src/audio/gain_node.rs
```diff
@@ -56,11 +59,6 @@ impl AudioNodeEngine for GainNode {
         }
         inputs
     }
-    fn message(&mut self, msg: AudioNodeMessage, sample_rate: f32) {
-        match msg {
-            AudioNodeMessage::SetAudioParamEvent(event) => {
-                self.gain.insert_event(event.to_event(sample_rate))
```
Manishearth: This reverts the AudioParam changes for gainnode

src/audio/node.rs
```diff
@@ -41,20 +41,26 @@ impl BlockInfo {
...
-
-    }
+    fn message(&mut self, _: AudioNodeMessage, _sample_rate: f32) {}
 }
 
-
 pub enum AudioNodeMessage {
-    SetAudioParamEvent(UserAutomationEvent)
+    OscillatorNode(OscillatorNodeMessage),
+    GainNode(GainNodeMessage),
```
Manishearth: I think we should be able to send all kinds of messages to all nodes, and nodes filter what they can actually handle.

Also, this reverts the AudioParam changes.

So you have SetAudioParam, Start, and Stop.

src/audio/gain_node.rs
ferjm: They are moved [here](https://github.com/servo/media/pull/31/files/3748b6b36b0ee84ed1dc7ba7e0655e41b84d88f2#diff-ab2867bdbfa223725105fd3200838b45R36)

Manishearth: Oh, I see what you did.

(sorry, reviewing on a phone)

I guess this is ok. A lot of the messages are common between nodes so I thought we'd have common messages. But this works.

Issue #23, opened 2018-05-23
Support for AudioContext state changes


ferjm:
We must expose a way to control and know the current [state](https://webaudio.github.io/web-audio-api/#enumdef-audiocontextstate) of the audio context.

ferjm:
Fixed by #32

PR #32, opened 2018-05-28
Support for BaseAudioContext.currentTime and state changes

### LINKED ISSUE #22: Implement support for the BaseAudioContext.currentTime property
[BaseAudioContext.currentTime](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-currenttime)

ferjm:
Fixes #22 and #23

### DIFF
--- a/src/audio/graph.rs
+++ b/src/audio/graph.rs
@@ -7,9 +7,17 @@ use std::thread::Builder;
 
 static NEXT_NODE_ID: AtomicUsize = ATOMIC_USIZE_INIT;
 
+#[derive(Debug, PartialEq)]
+pub enum ProcessingState {
+    Suspended,
+    Running,
+    Closed,
+}
+
 pub struct AudioGraph {
     sender: Sender<AudioRenderThreadMsg>,
     sync_sender: SyncSender<AudioRenderThreadSyncMsg>,
+    state: ProcessingState,
     sample_rate: f32,
 }
 
@@ -31,6 +39,7 @@ impl AudioGraph {
         Self {
             sender,
             sync_sender,
+            state: ProcessingState::Suspended,
             sample_rate,
         }
     }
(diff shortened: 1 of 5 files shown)


src/audio/graph.rs
```diff
@@ -2,26 +2,46 @@ use audio::node::{AudioNodeMessage, AudioNodeType};
...
 use std::sync::mpsc::{self, Sender};
+use std::sync::{Arc, RwLock};
 use std::thread::Builder;
 
 static NEXT_NODE_ID: AtomicUsize = ATOMIC_USIZE_INIT;
 
 pub struct AudioGraph {
     sender: Sender<AudioRenderThreadMsg>,
+    sample_rate: f32,
+    current_time: Arc<RwLock<f64>>,
```
Manishearth: does the AudioGraph need this? I'd rather have this be owned by the actual render thread without an RWLock

Manishearth: overall we should clean up the mutability story here but I have a plan for that

src/audio/render_thread.rs
```diff
@@ -117,7 +121,12 @@ impl AudioRenderThread {
                 }
                 // and push into the audio sink the result of processing a
                 // render quantum.
-                let _ = self.sink.push_data(self.process());
+                if let Ok(duration) = self.sink.push_data(self.process()) {
```
Manishearth: we shouldn't need to ask the sink for the duration; we know how long it is. The sink calculates the duration from information on the buffer that _we_ provided.

src/audio/graph.rs
ferjm: The WebAudio API exposes sync getters for these values, so I figured that we should make the control thread own them. I guess we can have a sync channels with buffer size 0 for these getters (?).

Manishearth:
Perhaps yeah.

Manishearth:
I don't think we need a sync sender here, we can use the same one -- just block on the receiver end.

PR #34, opened 2018-05-31
Implement basic AudioParams

Manishearth:
This implements the set and ramp AudioParam events and most of the framework for the harder ones.

r? @ferjm

### DIFF
--- a/examples/play.rs
+++ b/examples/play.rs
@@ -16,8 +16,11 @@ fn main() {
         assert_eq!(graph.current_time(), 0.);
         graph.resume();
         // change frequency at 0.5s and 1s
+        // change gain at 0.75s and 1.25s
         graph.message_node(0, AudioNodeMessage::SetAudioParamEvent(UserAutomationEvent::SetValueAtTime(110., 0.5)));
         graph.message_node(0, AudioNodeMessage::SetAudioParamEvent(UserAutomationEvent::SetValueAtTime(220., 1.)));
+        graph.message_node(1, AudioNodeMessage::SetAudioParamEvent(UserAutomationEvent::SetValueAtTime(0.25, 0.75)));
+        graph.message_node(1, AudioNodeMessage::SetAudioParamEvent(UserAutomationEvent::SetValueAtTime(0.75, 1.25)));
         thread::sleep(time::Duration::from_millis(2000));
         graph.suspend();
         thread::sleep(time::Duration::from_millis(500));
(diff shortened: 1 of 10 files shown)


src/audio/param.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        } else if let Some(next) = self.events.get(self.current_event + 1) {
+            if let Some(start_time) = next.start_time() {
+                if start_time >= current_tick {
+                    self.current_event += 1;
+                    self.event_start_time = current_tick;
+                    current_event = next;
+                }
+            }
+        }
```
ferjm: Should we remove old events from `self.events`?

src/audio/param.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Update the value of this param to the next
+    ///
+    /// Returns true if anything changed
+    pub fn update(&mut self, block: &BlockInfo, tick: Tick) -> bool {
+        if tick.0 != 0 && self.kind == ParamKind::KRate {
+            return false;
+        }
+
+        // println!("Curr {:?}", self.current_event);
```
ferjm: Delete this, please.

src/audio/param.rs
```diff
@@ -0,0 +1,173 @@
...
+        if self.events.len() <= self.current_event  {
+            return false;
+        }
+
+
+        let current_tick = block.absolute_tick(tick);
+        // println!("curr_tick {:?}", current_tick);
```
ferjm: Delete this one too, please.

examples/play.rs
```diff
@@ -14,14 +15,18 @@ fn main() {
         graph.create_node(AudioNodeType::GainNode(options));
         assert_eq!(graph.current_time(), 0.);
         graph.resume();
+        // change frequency at 0.5s and 1s
+        // change gain at 0.75s and 1.25s
+        graph.message_node(0, AudioNodeMessage::SetAudioParamEvent(UserAutomationEvent::SetValueAtTime(110., 0.5)));
+        graph.message_node(0, AudioNodeMessage::SetAudioParamEvent(UserAutomationEvent::SetValueAtTime(220., 1.)));
```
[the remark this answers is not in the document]
Manishearth: Yeah so the plan is to have an enum full of variants for each kind of param. I just didn't implement that here since it's trivial to do any time and we don't yet have anything with more than one param.

src/audio/param.rs
Manishearth: Eventually yeah. I want to wait till I implement the rest of the events because some of them have  _some_ lookbehind and I'm not clear on how that goes, exactly.

PR #36, opened 2018-06-01
Add SetTargetAtTime and Cancel* automation events

Manishearth:


Manishearth:
Only the value curve one is left over, and that's pretty straightforward.

The cancel ones need some work.

PR #38, opened 2018-06-04
Move audio sink into audio destination node 

ferjm:


ferjm:
I ended up storing the destination node separated from the graph. I think it'll be slightly simpler this way. Once we add graph support, we can simply store the indexes that are connected to the destination node input.

src/audio/destination_node.rs
```diff
@@ -1,16 +1,49 @@
...
+
+    pub fn has_enough_data(&self) -> bool {
+        self.sink.has_enough_data()
     }
 }
 
 impl AudioNodeEngine for DestinationNode {
-    fn process(&mut self, inputs: Chunk, _: &BlockInfo) -> Chunk {
-        inputs
+    fn process(&mut self, inputs: Chunk, _: &BlockInfo) -> Option<Chunk> {
```
Manishearth: I don't think we need to return Option here, Chunk::default is empty.

Also, can we switch back to the old sink model? My graph PR needs it to work since there's some extra work that may need to be done in between.

ferjm: Ok. Since the PR was approved and you didn't comment any further, I assumed you were ok with merging this. I'll revert it.

ferjm: 6131d39e6bdc9e300231c768a0b389e097c68fd9

Manishearth:
Yeah I just forgot to comment, that was my bad 😅😅

PR #45, opened 2018-06-06
Basic graph abstraction

Manishearth:
Todo: Using it in the code

r? @ferjm

### DIFF
--- a/src/audio/graph_impl.rs
+++ b/src/audio/graph_impl.rs
@@ -1,0 +1,97 @@
+use audio::block::Block;
+use petgraph::stable_graph::NodeIndex;
+use std::cell::RefCell;
+use audio::node::AudioNodeEngine;
+use audio::destination_node::DestinationNode;
+use petgraph::stable_graph::StableGraph;
+use petgraph::graph::DefaultIx;
+use petgraph::visit::EdgeRef;
+
+#[derive(Clone, Copy)]
+pub struct NodeId(pub usize);
+
+// we'll later alias NodeId to this
+pub type LocalNodeId = NodeIndex<DefaultIx>;
+
+/// A zero-indexed "port" for a node. Most nodes have one
+/// input and one output port, but some may have more
+///
+/// Kind is a zero sized type and is useful for distinguishing
+/// between input and output ports (which may otherwise share indices)
+pub type PortIndex<Kind> = (u32, Kind);
+
+/// An identifier for a port.
+pub type PortId<Kind> = (LocalNodeId, PortIndex<Kind>);
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct InputPort;
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct OutputPort;
+
+pub struct GraphImpl {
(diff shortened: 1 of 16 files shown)


Manishearth:
Can be merged as-is, opening it early so I can get feedback on the design before i integrate it with everything else.

src/audio/graph_impl.rs
```diff
@@ -0,0 +1,147 @@
...
+        // multiple edges between them between
+        // different ports. We should represent this somehow.
+        self.graph.add_edge(inp.0, out.0, Edge::new(inp.1, out.1));
+    }
+
+    pub fn dest_id(&self) -> LocalNodeId {
+        self.dest_id
+    }
+
+    pub fn process(&mut self, info: &BlockInfo) -> Chunk {
```
ferjm: This method could use some comments ;)

src/audio/graph_impl.rs
```diff
@@ -0,0 +1,147 @@
...
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct OutputPort;
+
+pub struct GraphImpl {
+    graph: StableGraph<Node, Edge>,
+    dest_id: LocalNodeId,
+}
+
+pub struct Node {
+    node: RefCell<Box<AudioNodeEngine>>,
```
ferjm: Note that AudioParams may be graph nodes as well.

src/audio/node.rs
```diff
@@ -46,6 +46,12 @@ pub trait AudioNodeEngine: Send {
     fn process(&mut self, inputs: Chunk, info: &BlockInfo) -> Chunk;
 
     fn message(&mut self, _: AudioNodeMessage, _sample_rate: f32) {}
+
+    fn input_count(&self) -> u32 { 1 }
+    fn output_count(&self) -> u32 { 1 }
+
+    /// If we're the destination node, extract the contained data
+    fn destination_data(&mut self) -> Option<Chunk> { None }
```
ferjm: Could we reuse `process` for this? With an extra flag in `BlockInfo` probably.

Manishearth: How? BlockInfo is immutable.

I don't want to do the model of the destination node pushing directly to the sink because the spec represents this as a separate "record the input for future retrieval" step. But we can still do that for now if you think it's better.

Manishearth:
For cycles: we can do the traversal a little bit more manually to support it, the algorithm isn't too hard. I was thinking once this is implemented we can parallelize out that work to support all the little graph things.

Manishearth:
For AudioParams the plan was to allow for params to "create" new input ports. We can have an additional boolean flag on PortId, and when flipped it enumerates params instead.

src/audio/node.rs
ferjm: Ok. I guess this is fine then.

I reverted 6131d39e6bdc9e300231c768a0b389e097c68fd9

ferjm:
I am going to go ahead and merge this as-is, as I am moving lots of code for the AudioBufferSourceNode work and this will likely cause conflicts.

Manishearth:
Sounds good. I'll do all the fixups when I get back from this hike.

PR #46, opened 2018-06-07
Basic AudioBufferSourceNode implementation

ferjm:
There are still several things to do for this node type but I thought I should push this as soon as possible as it is moving lots of code around.


src/audio/buffer_source_node.rs
```diff
@@ -0,0 +1,59 @@
...
+        inputs.blocks.push(Default::default());
+
+        if self.offset >= self.buffer.len() {
+            return inputs;
+        }
+
+        {
+            let data = inputs.blocks[0].data_mut();
+
+            for sample in data.iter_mut() {
```
Manishearth: We can just do a memcpy here I think?

https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.copy_from_slice

Manishearth:
I'm going to merge early so that we don't have too many conflicts, please look into the loop copy issue when you have time :smile:

Issue #43, opened 2018-06-05
Support for <audio>


ferjm:
This is not a high priority right now, but we need to start thinking about this.

We need to accommodate the current design to support playback for [HTMLMediaElement](https://html.spec.whatwg.org/multipage/media.html#htmlmediaelement)s. We can start with <audio> support for `<audio>`.

Some random thoughts about this:

- We need a different backend.
For WebAudio we have a [GStreamer src element](https://github.com/servo/media/blob/master/src/backends/gstreamer/audio_sink.rs#L20) that gets and deals with raw audio data.
For `<audio>` (and `<video>`) we probably need a different element to, for example, read from http:// URIs, which typically means raw unidentified data that could be encoded or not, could be a live stream or not, etc.
We may use [GstPlayer](https://gstconf.ubicast.tv/videos/gstplayer-a-simple-cross-platform-api-for-all-your-media-playback-needs-part-1/) for this, that does most part of the required work (via [playbin](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-base-plugins/html/gst-plugins-base-plugins-playbin.html)). This is what @philn and @sdroege used for their <video> [proof of concept](https://coaxion.net/blog/2017/10/rendering-html5-video-in-servo-with-gstreamer/).

- We need to support having `<audio>` elements as sources of WebAudio pipelines as well (check [MediaElementAudioSourceNode](https://webaudio.github.io/web-audio-api/#mediaelementaudiosourcenode)), which makes things a bit harder. I think we can instruct GstPlayer to output the processed stream into a custom [audio sink](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-base-plugins/html/gst-plugins-base-plugins-playbin.html#GstPlayBin--audio-sink) from where we can inject the resulting buffer into the WebAudio pipeline.

- We need to differentiate between real-time and non-real-time audio rendering and figure out in general how (or even if we want/need) to reuse the current WebAudio graph as a wider media graph to manage the different streams coming from different media APIs.

sdroege:
> I think we can instruct GstPlayer to output the processed stream into a custom audio sink from where we can inject the resulting buffer into the WebAudio pipeline.

Yes via an `appsink` :)

ferjm:
HTMLMediaElements playback is already supported via GstPlayer.

PR #52, opened 2018-06-13
Audio decoder

### LINKED ISSUE #41: Support for BaseAudioContext.decodeAudioData
The decoding work needs to happen on a dedicated thread. We can probably use GStreamer [decodebin](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-base-plugins/html/gst-plugins-base-plugins-decodebin.html) for this.

ferjm:
Fixes #41

### DIFF
--- a/examples/audio_decoder.rs
+++ b/examples/audio_decoder.rs
@@ -1,15 +1,14 @@
 extern crate servo_media;
 
 use servo_media::audio::buffer_source_node::AudioBufferSourceNodeMessage;
-use servo_media::audio::decoder::AudioDecoderMsg;
+use servo_media::audio::decoder::AudioDecoderCallbacks;
 use servo_media::audio::node::{AudioNodeMessage, AudioNodeType};
 use servo_media::ServoMedia;
 use std::env;
 use std::fs::File;
 use std::io::Read;
 use std::sync::mpsc;
-use std::sync::Arc;
-use std::thread::Builder;
+use std::sync::{Arc, Mutex};
 use std::{thread, time};
 
 fn run_example(servo_media: Arc<ServoMedia>) {
@@ -23,29 +22,23 @@ fn run_example(servo_media: Arc<ServoMedia>) {
     let mut file = File::open(filename).unwrap();
     let mut bytes = vec![];
     file.read_to_end(&mut bytes).unwrap();
+    let decoded_audio = Arc::new(Mutex::new(Vec::new()));
+    let progress = decoded_audio.clone();
     let (sender, receiver) = mpsc::channel();
-    graph.decode_audio_data(bytes.to_vec(), sender);
-    let (sender2, receiver2) = mpsc::channel();
-    Builder::new()
-        .name("AudioDecoder receiver".to_owned())
-        .spawn(move || {
-            let mut decoded_audio = Vec::new();
-            loop {
-                match receiver.recv().unwrap() {
-                    AudioDecoderMsg::Eos => {
-                        let _ = sender2.send(decoded_audio);
-                        break;
-                    }
-                    AudioDecoderMsg::Error => break,
-                    AudioDecoderMsg::Progress(progress) => {
-                        decoded_audio.extend_from_slice(&progress);
-                    }
-                }
-            }
+    let callbacks = AudioDecoderCallbacks::new()
+        .eos(move || {
+            sender.send(()).unwrap();
         })
-        .unwrap();
-    println!("Decoding");
-    let decoded_audio = receiver2.recv().unwrap();
+        .error(|| {
+            eprintln!("Error decoding audio");
+        })
+        .progress(move |buffer| {
+            progress.lock().unwrap().extend_from_slice(&buffer);
+        })
+        .build();
+    graph.decode_audio_data(bytes.to_vec(), callbacks);
+    println!("Decoding audio");
+    receiver.recv().unwrap();
     println!("Audio decoded");
     let buffer_source = graph.create_node(AudioNodeType::AudioBufferSourceNode);
     let dest = graph.dest_node();
@@ -57,7 +50,7 @@ fn run_example(servo_media: Arc<ServoMedia>) {
     graph.message_node(
         buffer_source,
         AudioNodeMessage::AudioBufferSourceNode(AudioBufferSourceNodeMessage::SetBuffer(
-            decoded_audio,
+            decoded_audio.lock().unwrap().to_vec(),
         )),
     );
     graph.resume();
--- a/examples/audio_decoder.rs
+++ b/examples/audio_decoder.rs
@@ -1,10 +1,15 @@
 extern crate servo_media;
 
+use servo_media::audio::buffer_source_node::AudioBufferSourceNodeMessage;
+use servo_media::audio::decoder::AudioDecoderMsg;
+use servo_media::audio::node::{AudioNodeMessage, AudioNodeType};
 use servo_media::ServoMedia;
 use std::env;
 use std::fs::File;
 use std::io::Read;
+use std::sync::mpsc;
 use std::sync::Arc;
+use std::thread::Builder;
 use std::{thread, time};
 
 fn run_example(servo_media: Arc<ServoMedia>) {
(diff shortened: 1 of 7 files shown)


servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -0,0 +1,142 @@
...
+
+                let sink_pad = queue.get_static_pad("sink").expect("queue has no sinkpad");
+                src_pad
+                    .link(&sink_pad)
+                    .into_result()
+                    .map(|_| ())
+                    .map_err(|_| ())
+            };
+
+            if !is_audio || insert_sink().is_err() {
```
sdroege: You probably also do not want to add a new sink for each audio stream (there could be multiple), but only take one. Or do you want to take them all?

servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -0,0 +1,142 @@
...
+                // XXX reject Future.
+                return;
+            }
+        });
+
+        let appsrc = appsrc.downcast::<AppSrc>().map_err(|_| ())?;
+        appsrc.set_property_format(gst::Format::Time);
```
sdroege: Time format requires you to properly timestamp the data you put into the appsrc for each buffer. You probably can't do that, so set it to bytes instead :)

servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -0,0 +1,142 @@
...
+                        let data = data.as_mut_slice();
+                        buffer
+                            .copy_from_slice(0, &data[offset..next_offset])
+                            .expect("copying failed");
+                        offset = next_offset;
+                    }
+                    appsrc_
+                        .push_buffer(buffer)
+                        .into_result()
+                        .expect("pushing buffer failed");
```
sdroege: This should be cleanly handled instead of just panicking

servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -0,0 +1,142 @@
...
+                    };
+                    let mut buffer = gst::Buffer::with_size(buffer_size).unwrap();
+                    {
+                        let buffer = buffer.get_mut().unwrap();
+                        let next_offset = offset + buffer_size;
+                        let data = data.as_mut_slice();
+                        buffer
+                            .copy_from_slice(0, &data[offset..next_offset])
+                            .expect("copying failed");
+                        offset = next_offset;
```
sdroege: You could use `std::io::Cursor` and `Read` on the `Vec` to prevent having to keep track of the offset yourself

ferjm:
With these patches I am able to decode and play an .mp3 file, but with very poor quality (lots of noise). I am not sure why yet.

philn:
Seems like you need a capsfilter before the appsink so that the audio samples are converted to the right format expected by the backend.

sdroege:
> Seems like you need a capsfilter before the appsink so that the audio samples are converted to the right format expected by the backend.

Or you set the caps property on appsink accordingly. That has the same effect and is fewer elements

examples/audio_decoder.rs
```diff
@@ -0,0 +1,70 @@
...
+        AudioNodeMessage::AudioBufferSourceNode(AudioBufferSourceNodeMessage::Start(0.)),
+    );
+    graph.message_node(
+        buffer_source,
+        AudioNodeMessage::AudioBufferSourceNode(AudioBufferSourceNodeMessage::SetBuffer(
+            decoded_audio.lock().unwrap().to_vec(),
+        )),
+    );
+    graph.resume();
+    thread::sleep(time::Duration::from_millis(5000));
```
sdroege: Why?

ferjm: Because if the graph is dropped, the rendering thread is closed and we won't hear the audio playback.

ferjm:
With the appsink caps set, the sound is perfect. Thank you!

servo-media/src/audio/graph.rs
```diff
@@ -90,7 +93,22 @@ impl AudioGraph {
...
 
     pub fn connect_ports(&self, from: PortId<OutputPort>, to: PortId<InputPort>) {
-        let _ = self.sender.send(AudioRenderThreadMsg::ConnectPorts(from, to));
+        let _ = self.sender
+            .send(AudioRenderThreadMsg::ConnectPorts(from, to));
+    }
+
+    /// Asynchronously decodes the audio file data contained in the given
+    /// buffer.
+    pub fn decode_audio_data(&self, data: Vec<u8>, callbacks: AudioDecoderCallbacks) {
```
Manishearth: We should _eventually_ split out the GStreamer stuff as a separate crate with traits, but that's going to be annoying for now so we don't need to do it yet 😄 

(it's also low priority)

examples/audio_decoder.rs
```diff
@@ -0,0 +1,70 @@
...
+            eprintln!("Error decoding audio");
+        })
+        .progress(move |buffer| {
+            progress
+                .lock()
+                .unwrap()
+                .extend_from_slice((*buffer).as_ref());
+        })
+        .build();
+    graph.decode_audio_data(bytes.to_vec(), callbacks);
```
Manishearth: Should we eventually make it such that this can be plugged into a source element on the render thread side, or do we want this to be separate to make the network stuff work?

examples/audio_decoder.rs
```diff
@@ -0,0 +1,70 @@
...
+    println!("Decoding audio");
+    receiver.recv().unwrap();
+    println!("Audio decoded");
+    let buffer_source = graph.create_node(AudioNodeType::AudioBufferSourceNode);
```
Manishearth: How do we plan to make this work with ongoing network fetches? Perhaps a BufferSourceNode you can message extra buffers to?

(not something we need to solve now, just want your thoughts)

servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -0,0 +1,195 @@
...
+                // XXX Issue #18 channels support.
+                let audio_info = gst_audio::AudioInfo::new(gst_audio::AUDIO_FORMAT_F32, 44100, 1)
+                    .build()
+                    .ok_or(())?;
+                appsink.set_caps(&audio_info.to_caps().unwrap());
+
+                let pipeline_ = pipeline.clone();
+                let pipeline__ = pipeline.clone();
+                let callbacks_ = callbacks.clone();
+                let callbacks__ = callbacks.clone();
```
Manishearth: not too happy about this but this seems to make sense 😄

servo-media/src/audio/graph.rs
ferjm: We already have the `AudioSink` and `AudioDecoder` traits, so it shouldn't be too hard ™ :)

examples/audio_decoder.rs
ferjm: For media elements (and MediaElementAudioSourceNodes) we will likely use a different approach and use `GstPlayer` directly (which handles all the decoding work internally), check https://github.com/servo/media/issues/43.

For MediaStream*AudioSourceNodes, I am not so sure yet, I still need to read more of these specs, but in general, I have the feeling that all decoding should happen outside of any WebAudio source element, and certainly off the render thread.

In any case, `AudioDecoder` and `GStreamerAudioDecoder` are meant to be reusable.

examples/audio_decoder.rs
ferjm: AFAICT `AudioBufferSourceNodes` are meant for in-memory audio assets. For network-backed assets the spec suggests the usage of `AudioWorkletNodes`, which is a completely different war :).

servo-media/src/backends/gstreamer/audio_decoder.rs
ferjm: Yeah, I am not too happy either, but I couldn't think of a better approach :\

ferjm:
Thanks for the review and feedback, folks!

examples/audio_decoder.rs
```diff
@@ -0,0 +1,70 @@
...
+    let progress = decoded_audio.clone();
+    let (sender, receiver) = mpsc::channel();
+    let callbacks = AudioDecoderCallbacks::new()
+        .eos(move || {
+            sender.send(()).unwrap();
+        })
+        .error(|| {
+            eprintln!("Error decoding audio");
+        })
+        .progress(move |buffer| {
```
sdroege: Is there another user for this already apart from the example? Some code that does not actually copy all the samples directly but reuses the slice?

servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -0,0 +1,195 @@
...
+
+            let insert_sink = || -> Result<(), ()> {
+                let queue = gst::ElementFactory::make("queue", None).ok_or(())?;
+                let convert = gst::ElementFactory::make("audioconvert", None).ok_or(())?;
+                let resample = gst::ElementFactory::make("audioresample", None).ok_or(())?;
+                let sink = gst::ElementFactory::make("appsink", None).ok_or(())?;
+                let appsink = sink.clone().dynamic_cast::<AppSink>().map_err(|_| ())?;
+
+                // XXX Issue #18 channels support.
+                let audio_info = gst_audio::AudioInfo::new(gst_audio::AUDIO_FORMAT_F32, 44100, 1)
```
sdroege: If 44100 Hz are not required, 48000 might be nicer... or can the sample rate even be signalled somewhere via webaudio so you wouldn't have to constrain it here?

examples/audio_decoder.rs
ferjm: Not yet. I'm working on the DOM bindings using this code.

servo-media/src/backends/gstreamer/audio_decoder.rs
ferjm: Yeah, the sample rate should probably be given by the AudioContext. I'll write a patch for this.

sdroege: You then want to set the caps similar to this (and possibly later also allow a different number of channels): https://github.com/sdroege/gstreamer-rs/blob/db3fe694154c697afdaf3efb6ec65332546942e0/examples/src/bin/appsink.rs#L54-L62

PR #53, opened 2018-06-20
Minor decoder fixes

ferjm:
Addressing this review feedback https://github.com/servo/media/pull/52#discussion_r196360133

### DIFF
--- a/servo-media/src/backends/gstreamer/audio_decoder.rs
+++ b/servo-media/src/backends/gstreamer/audio_decoder.rs
@@ -1,6 +1,6 @@
 use super::gst_app::{AppSink, AppSinkCallbacks, AppSrc};
 use super::gst_audio;
-use audio::decoder::{AudioDecoder, AudioDecoderCallbacks};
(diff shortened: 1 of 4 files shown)


servo-media/src/backends/gstreamer/audio_decoder.rs
```diff
@@ -97,11 +104,15 @@ impl AudioDecoder for GStreamerAudioDecoder {
...
-                    .build()
-                    .ok_or(())?;
-                appsink.set_caps(&audio_info.to_caps().unwrap());
+                appsink.set_caps(&gst::Caps::new_simple(
+                    "audio/x-raw",
+                    &[
+                        ("format", &gst_audio::AUDIO_FORMAT_F32.to_string()),
+                        ("layout", &"interleaved"),
+                        ("channels", &(options.channels as i32)),
+                        ("rate", &(options.sample_rate as i32)),
```
sdroege: Well if you only want to handle one specific number of channels/rate here you can and probably should continue using `AudioInfo`. The full caps API is only needed here if you want to support e.g. ranges of channels (say, 1-16 channels) or rates

Issue #18, opened 2018-05-22
Channel support


Manishearth:
With #17 we have a chunk and block abstraction, but the block abstraction only supports a single channel.

We should expand this to allow for more channels (stored in the same buffer, probably), as well as upmixing/downmixing.

sdroege:
From the GStreamer side, all channels for a sample need to be in the same buffer. And ideally interleaved (e.g. for stereo you have `lrlrlrlr...`). Proper support for planar audio is currently being [added](https://bugzilla.gnome.org/show_bug.cgi?id=751605)

sdroege:
Ok, then you would ideally convert between planar and interleaved at the boundary to GStreamer for now. Simple enough to do :)

Proper planar support should be in GStreamer with the next stable release in about 4-5 months, but you probably don't want to directly depend on such a new version.

Manishearth:
Yeah, that's the plan!

Manishearth:
@sdroege So how should I be sending audio with multiple channels to our GStreamer AppSrc? And how do I obtain the number of channels supported here?

Manishearth:
Figured it out!

sdroege:
All good then? :)

ferjm:
@Manishearth Can we close this one? IIUC the only remaining work to have channels support is #55

PR #54, opened 2018-06-21
Add basic channels abstraction, ChannelMergerNode, and sink support for channels

Manishearth:
- Extends Block to support channels.
 - Makes the DestinationNode have two channels by default. We'll later make this something that the user can set.
 - Adds ChannelMergerNode
 - Implements _very basic_ upmixing from one channel to multiple. Proper upmixing will come later.

Channels are stored as a continuous buffer with each FRAMES_PER_BLOCK chunk being one channel. They can also be grouped together  via `repeat` -- e.g. if an oscillator node produces 2 channels of output, it will actually produce one channel with a `repeat` bit set. Similarly, one-channel blocks will be upmixed to 2-channel `repeat` blocks.

The iterator is a bit awkward due to the lack of associated type constructors (it's a streaming iterator).

part of #18

### DIFF
--- a/examples/channels.rs
+++ b/examples/channels.rs
@@ -1,0 +1,45 @@
+extern crate servo_media;
+
+use servo_media::audio::channel_node::ChannelNodeOptions;
+use servo_media::audio::gain_node::GainNodeOptions;
+use servo_media::audio::node::{AudioNodeMessage, AudioNodeType};
+use servo_media::audio::oscillator_node::OscillatorNodeMessage;
+use servo_media::audio::param::UserAutomationEvent;
+use servo_media::ServoMedia;
+use std::{thread, time};
+
+fn main() {
+    if let Ok(servo_media) = ServoMedia::get() {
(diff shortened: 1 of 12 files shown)


Manishearth:
We can land this now before everything else is in place just to make things easier

Manishearth:
Scratch that, I planned ahead pretty well and the actual stuff was easy to implement with this abstraction in place.

examples/channels.rs
```diff
@@ -0,0 +1,45 @@
+extern crate servo_media;
+
+use servo_media::audio::channel_node::ChannelNodeOptions;
+use servo_media::audio::gain_node::GainNodeOptions;
+use servo_media::audio::node::{AudioNodeMessage, AudioNodeType};
+use servo_media::audio::oscillator_node::OscillatorNodeMessage;
+use servo_media::audio::param::UserAutomationEvent;
```
ferjm: This is not used.

PR #56, opened 2018-06-22
Some graph API changes

ferjm:
I identified these needs while working on the DOM bindings.

<bikeshedding>I am tempted to rename AudioGraph to AudioContext everywhere. wdyt?</bikeshedding>


Manishearth:
Yeah naming it to context is fine. That way we can call GraphImpl AudioGraph

Manishearth:
merging, fix the options and context thing if you want 😄

PR #58, opened 2018-06-26
Channel fixes

Manishearth:
- Adds channel count modes
 - Adds channel interpretation modes
 - Adds all the upmix/downmix equations

Haven't yet written tests for this partially because I don't have quad/5.1 systems to test with.

r? @ferjm

### DIFF
--- a/servo-media/src/audio/block.rs
+++ b/servo-media/src/audio/block.rs
@@ -129,7 +129,13 @@ impl Block {
     }
 
     pub fn data_chan_frame(&self, frame: usize, chan: u8) -> f32 {
-        self.buffer[frame + chan as usize * FRAMES_PER_BLOCK_USIZE]
+        let offset = if self.repeat {
+            0
+        } else {
+            chan as usize * FRAMES_PER_BLOCK_USIZE
+        };
+
+        self.buffer[frame + offset]
     }
 
     /// upmix/downmix the channels if necessary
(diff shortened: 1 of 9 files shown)


servo-media/src/backends/gstreamer/audio_sink.rs
```diff
@@ -124,7 +124,7 @@ impl AudioSink for GStreamerAudioSink {
             // sometimes nothing reaches the output
             if chunk.len() == 0 {
                 chunk.blocks.push(Default::default());
-                chunk.blocks[0].mix(channels as u8);
+                chunk.blocks[0].repeat(channels as u8);
```
sdroege: You probably want to also signal the actual channel layout to GStreamer via the channel-mask or the positions array in the GstAudioInfo

servo-media/src/audio/block.rs
```diff
@@ -126,15 +128,232 @@ impl Block {
...
+            // If we're discrete, have a repeat, and are downmixing,
+            // just truncate by changing the channel value
+            if self.repeat && self.channels > channels {
+                self.channels = channels;
+            } else {
+                // otherwise resize the buffer, silent-filling when necessary
+                self.resize_silence(channels);
+            }
+        } else {
+            // For speakers, we have to do special things based the
```
ferjm: typo: "based on the"

servo-media/src/backends/gstreamer/audio_sink.rs
Manishearth: We only support one layout in gstreamer right now, yeah?

Manishearth: Oh, I see. I'll be working on this when I add better support for output channel counts. Currently it's hardcoded to be two.

PR #67, opened 2018-07-02
Reduce latency by setting appsrc max-bytes to the block size

ferjm:
With this patch I can notice a much lower latency running the theremin demo.

r? @Manishearth @sdroege

### DIFF
--- a/servo-media/src/backends/gstreamer/audio_sink.rs
+++ b/servo-media/src/backends/gstreamer/audio_sink.rs
@@ -76,6 +76,9 @@ impl AudioSink for GStreamerAudioSink {
         self.set_audio_info(sample_rate, 2)?;
         self.appsrc.set_property_format(gst::Format::Time);
 
+        // Allow only a single chunk.
+        self.appsrc.set_max_bytes(1);
+
         let appsrc = self.appsrc.clone();
         Builder::new()
             .name("GstAppSrcCallbacks".to_owned())


servo-media/src/backends/gstreamer/audio_sink.rs
```diff
@@ -76,6 +76,13 @@ impl AudioSink for GStreamerAudioSink {
         self.set_audio_info(sample_rate, 2)?;
         self.appsrc.set_property_format(gst::Format::Time);
 
+        let audio_info = self.audio_info.borrow();
+        let audio_info = audio_info.as_ref().unwrap();
+        let bpf = audio_info.bpf() as usize;
+        let n_samples = FRAMES_PER_BLOCK.0 as u64;
+        let buf_size = (n_samples as usize) * (bpf);
+        self.appsrc.set_max_bytes(buf_size as u64);
```
sdroege: This makes no sense for reducing latency :) It would mean that any additional data is dropped (see `block` property) instead of being queued up. Which would mean that you either should implement some kind of backpressure here or you're just producing too much data to begin with.

This needs some closer investigation why it has an effect at all. The appsrc is forwarding data as fast as it can (not once the limit is reached), and downstream is consuming data in real-time.

ferjm: We have a backpressure mechanism [here](https://github.com/ferjm/media/blob/a95e063729324c359976236104d825244bb180e8/servo-media/src/audio/render_thread.rs#L133). We stop feeding the appsrc with data if we reached `max_bytes` (which, after this patch, will happen every time we push a single rendered block into the appsrc). IIUC this would essentially be the same as setting the `block` property on. But IIRC we couldn't use the `block` property because we wanted to continue processing messages sent to the render thread while we are waiting for the sink to process the audio.

sdroege: Ah I see. You could as well set `max-bytes` to 1 if you only ever want to allow a single chunk btw :)

Makes sense then, I was forgetting that context.

Issue #68, opened 2018-07-03
Support fan-in, fan-out, multiple connections between nodes


Manishearth:
Currently we incorrectly upmix/downmix across ports instead of mixing multiple inputs to a single port.

We also don't support fan-in and fan-out for ports at all, and don't support multiple connections between nodes.

We need to:

 - Modify `Edge` so it contains a smallvec of port pairs. Modify the graph code to support this well
 - Modify the processing code to clone chunks in case of fan-out
 - Modify the caching code to store a smallvec of chunks
 - Modify the processing code to mix these chunks together (and ultimately sum them)

Manishearth:
While doing this we should also fix the direction of the edges via the `Reversed` adaptor so that it's not incredibly confusing.

(Currently the edges go in the opposite direction they should, which is unfortunate)

PR #69, opened 2018-07-03
Invert edge direction

Manishearth:
(part of #68)

We currently have the edges going in the reverse direction as the audio,
which is confusing. The reason we did this was so that we could use the
DFSPostOrder iterator.

This inverts their direction so that edges now go from source to sink, and
uses a `Reversed` adaptor to fix it.

r? @ferjm

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[alias]
+ex = "run -p examples --bin"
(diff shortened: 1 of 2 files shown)


.cargo/config
```diff
@@ -0,0 +1,2 @@
+[alias]
+ex = "run -p examples --bin"
```
ferjm: It would be nice to update the README section about running the examples to mention the existence of this alias.

PR #71, opened 2018-07-05
Allow sharing AudioBufferSourceNode buffer

ferjm:
r? @Manishearth

### DIFF
--- a/examples/audio_decoder.rs
+++ b/examples/audio_decoder.rs
@@ -1,6 +1,6 @@
 extern crate servo_media;
 
-use servo_media::audio::buffer_source_node::AudioBufferSourceNodeMessage;
+use servo_media::audio::buffer_source_node::{AudioBuffer, AudioBufferSourceNodeMessage};
 use servo_media::audio::decoder::AudioDecoderCallbacks;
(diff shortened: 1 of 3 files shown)


ferjm:
I am not sure if locking here is preferred over copying (in general, what I hear for real-time audio is that while copying may take longer, being deterministic makes it preferable over locking), but since we talked about sharing the buffer, I'm proposing this patch.

The DOM bindings using this are [here](https://github.com/ferjm/servo/blob/webaudio/components/script/dom/audiobuffer.rs#L28) and [here](https://github.com/ferjm/servo/blob/webaudio/components/script/dom/audiobuffersourcenode.rs#L116)

Manishearth:
I don't think we should be locking here; for one there will be a process boundary here which makes locks hard to do, and for the other this means that not only can audio code block DOM (it can already), DOM code can now block audio. The way I thought we could share it would be to make it possible to send messages to the thread asking for buffer changes and asking for the buffer contents. But this has its own problems.

I suspect the realtime requirements for uploading buffers are more relaxed, though, so maybe just copying is the best.

Feel free to merge -- I don't have an objection to this PR; just not sure if it's the best model.

ferjm:
We'll go with the copy approach.

PR #78, opened 2018-07-09
Unify setparam methods, various cleanups

Manishearth:
This adds a ParamType that lets us talk about AudioParams in a unified way. From the DOM side an `AudioParam` can now just be a `(NodeId, ParamType)` pair.

In the future we can use this to talk of connecting nodes to audio params as well, as an variant of PortId.

I also did some cleanups, marking engine-related things crate-public (script should never create engines), and renaming `AudioNodeType` to `AudioNodeInit` (this will require a rename in the DOM code) to better reflect what it does. `AudioNodeType` now exists as a bare enum for debugging purposes.

r? @ferjm


Manishearth:
Servo side at https://github.com/ferjm/servo/pull/4

PR #84, opened 2018-07-11
Split up crates to facilitate backend switching

Manishearth:
This gets rid of the gst feature, instead using target-switching to select the backend.

The servo-media crate is a thin wrapper that does all the backend switching. The GStreamer backend is 
now in the servo-media-gstreamer crate, and all of the audio stuff is in the servo-media-audio crate.

You use the crate the same way (one small difference, in accordance to rust style you will have to import the crate as `servo-media` in Cargo.toml), and it automatically picks the backend based on the platform.

This makes it pretty easy to swap backends as well; adding a new backend can be done with a cargo feature, and once we fix the android stuff we can make this unconditional.

r? @ferjm


servo-media-audio/src/node.rs.orig
```diff
@@ -0,0 +1,225 @@
+<<<<<<< HEAD:servo-media-audio/src/node.rs
```
CYBAI: Maybe this one is commited accidentally?

Manishearth:
I had to git merge just because rebasing was throwing up way too many merge conflicts.

servo-media-audio/src/node.rs.orig
Manishearth: fixed

Manishearth:
I'd kinda mentioned this last week but I didn't really explain it well 😄

PR #87, opened 2018-07-16
Add support for the AudioScheduledSourceNode.ended event

ferjm:
https://webaudio.github.io/web-audio-api/#dom-audioscheduledsourcenode-onended

r? @Manishearth


Manishearth:
r=me after rebase.

I think we can get rid of the mutex here, we're passing by ownership.

Another solution is to let "setcallback" accept a sender that script can listen on (but this requires some extra plumbing to make work)

PR #88, opened 2018-07-17
Add Player as a crate

ceyusa:
This pull request is rebased above #87

### DIFF
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -1,0 +1,384 @@
+use glib::*;
+use gst;
+use gst_app;
+use gst_player;
+use gst_player::{PlayerMediaInfo, PlayerStreamInfoExt};
+use servo_media_player::frame::{Frame, FrameRenderer};
+use servo_media_player::metadata::Metadata;
+use servo_media_player::{PlaybackState, Player, PlayerEvent};
+use std::sync::mpsc::{self, Sender};
+use std::sync::{Arc, Mutex};
+use std::time;
+use std::u64;
+
+fn frame_from_sample(sample: &gst::Sample) -> Result<Frame, ()> {
+    let caps = sample.get_caps().ok_or_else(|| ())?;
+    let s = caps.get_structure(0).ok_or_else(|| ())?;
+    let width = s.get("width").ok_or_else(|| ())?;
+    let height = s.get("height").ok_or_else(|| ())?;
+
+    let buffer = sample.get_buffer().ok_or_else(|| ())?;
+    let map = buffer.map_readable().ok_or_else(|| ())?;
+    let data = Vec::from(map.as_slice());
+
+    Ok(Frame::new(width, height, Arc::new(data)))
+}
+
+fn metadata_from_media_info(media_info: &PlayerMediaInfo) -> Result<Metadata, ()> {
+    let dur = media_info.get_duration();
+    let duration = if dur != gst::ClockTime::none() {
+        let mut nanos = dur.nanoseconds().ok_or_else(|| ())?;
+        nanos = nanos % 1_000_000_000;
+        let seconds = dur.seconds().ok_or_else(|| ())?;
+        Some(time::Duration::new(seconds, nanos as u32))
+    } else {
+        None
+    };
+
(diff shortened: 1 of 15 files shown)


backends/gstreamer/src/player.rs
```diff
@@ -0,0 +1,397 @@
...
+        self.inner
+            .lock()
+            .unwrap()
+            .register_frame_renderer(renderer);
+    }
+
+    fn set_input_size(&self, size: u64) {
+        if let Ok(mut inner) = self.inner.lock() {
+            inner.set_input_size(size);
+        }
```
ferjm:
```rust
self.inner.lock().unwrap().set_input_size(size);
```

backends/gstreamer/src/player.rs
```diff
@@ -0,0 +1,397 @@
...
+
+    pub fn stop(&mut self) {
+        self.player.stop();
+        self.last_metadata = None;
+        self.appsrc = None;
+    }
+
+    pub fn start(&mut self) {
+        self.player.pause();
+    }
```
ferjm: This is a bit confusing :) (start vs pause)

player/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn register_event_handler(&self, sender: ipc::IpcSender<PlayerEvent>);
+    fn register_frame_renderer(&self, renderer: Box<frame::FrameRenderer>);
+
+    fn setup(&self) -> bool;
+    fn play(&self);
+    fn stop(&self);
+
+    fn set_input_size(&self, size: u64);
+    fn push_data(&self, data: Vec<u8>) -> bool;
+    fn end_of_stream(&self) -> bool;
```
ferjm: `Result<(), ()>` as the return value of `setup`, `push_data` and `end_of_stream` is probably more idiomatic and let us extend it with specific errors in the future if needed.

examples/player.rs
```diff
@@ -0,0 +1,107 @@
...
+    player.lock().unwrap().stop();
+}
+
+fn main() {
+    if let Ok(servo_media) = ServoMedia::get() {
+        run_example(servo_media);
+    } else {
+        unreachable!()
+    }
+}
```
ferjm: Not required for this PR, but it would be really nice if we could do something like [gecko-media's test-player](https://github.com/servo/gecko-media/blob/master/examples/src/bin/test-player.rs) :-)

backends/gstreamer/src/player.rs
```diff
@@ -0,0 +1,397 @@
...
+use std::string;
+use std::sync::mpsc;
+use std::sync::{Arc, Mutex};
+use std::time;
+use std::u64;
+
+struct GStreamerFrame {}
+
+impl GStreamerFrame {
+    fn new(sample: &gst::Sample) -> Frame {
```
Manishearth: This can just be a function, i think, no need for the GStreamerFrame type

backends/gstreamer/src/player.rs
```diff
@@ -0,0 +1,397 @@
...
+        let buffer = sample.get_buffer().unwrap();
+
+        let map = buffer.map_readable().unwrap();
+        let data = Vec::from(map.as_slice());
+
+        Frame::new(width, height, Arc::new(data))
+    }
+}
+
+struct GStreamerMetadata {}
```
Manishearth: ditto, this should just be a function like `create_metadata()`

Manishearth: Alternatively, this can be a `From` impl, `impl From<&PlayerMediaInfo> for Metadata`

backends/gstreamer/src/player.rs
```diff
@@ -0,0 +1,397 @@
...
+        }
+    }
+}
+
+struct PlayerInner {
+    player: gst_player::Player,
+    appsrc: Option<gst_app::AppSrc>,
+    appsink: gst_app::AppSink,
+    input_size: u64,
+    subscribers: Vec<ipc::IpcSender<PlayerEvent>>,
```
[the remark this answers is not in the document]
Manishearth: yes, I'd use MPSC.

We can later drop in the process boundary

PR #96, opened 2018-07-26
Turn off gstreamer for non-x86

Manishearth:
We don't have the ability to include it as a dependency on our CI right now.

### DIFF
--- a/servo-media/Cargo.toml
+++ b/servo-media/Cargo.toml
@@ -12,5 +12,5 @@ path = "../audio"
 [dependencies.servo-media-player]
 path = "../player"
 
-[target.'cfg(not(target_os = "android"))'.dependencies.servo-media-gstreamer]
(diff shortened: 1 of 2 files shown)


sdroege:
Is the problem only the gstreamer binaries? If so, I'm planning to add CI for arm at least soon. What other targets do you build on your CI?

Manishearth:
Only for the gstreamer binaries.

linux-x86, aarch64, and arm, mac-x86, windows-x86, android-arm

I'm doing this temporarily so we can land stuff :)

sdroege:
> Only for the gstreamer binaries. linux-x86, aarch64, and arm, mac-x86, windows-x86, android-arm

Ah, Windows/mac/Android should be fine with the binaries that exist. For Linux I was planning to do armv7 for gstreamer-rs so you would have binaries for that available already then (just not sure when, quite busy currently). I could write you down how I build these binaries so you could do the same for linux/x86 and linux/aarch64 then

PR #102, opened 2018-08-01
Test player

ferjm:


ferjm:
r? @Manishearth 

The video is not centered or fullscreen yet, but I think this is enough to test for now. 

testplayer

(The frame rate is nicer IRL than on the gif)

PR #106, opened 2018-08-13
fixes, init and change to use ipc crate

ceyusa:
the first two commits are simple fixes to the code.

4bfa467 moves the init() function into a shared one between the audio and the player. This is a RFC, since perhaps it  can be defined as a trait. I'm not sure

7d28b6a it is required to pass the event messages through the ROUTER in servo


jdm:
These changes all look fine to me.

PR #109, opened 2018-08-23
Remove unused num-traits and fix lock

ferjm:
This completely remove an unused dependency from backends/gstreamer and fixes this build error:

```
error: failed to parse lock file at: /Users/ferjm/dev/mozilla/media/Cargo.lock

Caused by:
  package `num-traits 0.1.43 (registry+https://github.com/rust-lang/crates.io-index)` is specified as a dependency, but is missing from the package list
```

r? @ceyusa


ceyusa:
lgtm

PR #117, opened 2018-08-28
Deinterleave decoded audio

### LINKED ISSUE #55: Deinterleave decoded audio
With #54 in place, we can now deinterleave the decoded audio and create the AudioBuffer with the correct number of channels.

We can use [GStreamer's deinterleave](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-good-plugins/html/gst-plugins-good-plugins-deinterleave.html) element for this.

ferjm:
Fixes #55

### DIFF
--- a/audio/src/decoder.rs
+++ b/audio/src/decoder.rs
@@ -4,7 +4,8 @@ use std::sync::Mutex;
 pub struct AudioDecoderCallbacks {
     pub eos: Mutex<Option<Box<FnBox() + Send + 'static>>>,
     pub error: Mutex<Option<Box<FnBox() + Send + 'static>>>,
-    pub progress: Option<Box<Fn(Box<AsRef<[f32]>>, u32, u32) + Send + Sync + 'static>>,
+    pub progress: Option<Box<Fn(Box<AsRef<[f32]>>, u32) + Send + Sync + 'static>>,
+    pub ready: Mutex<Option<Box<FnBox(u32) + Send + 'static>>>,
 }
 
 impl AudioDecoderCallbacks {
@@ -13,6 +14,7 @@ impl AudioDecoderCallbacks {
             eos: None,
             error: None,
             progress: None,
+            ready: None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


backends/gstreamer/src/audio_decoder.rs
```diff
@@ -67,12 +67,30 @@ impl AudioDecoder for GStreamerAudioDecoder {
...
+        // Initial pipeline looks like
+        //
+        // appsrc ! decodebin2! ...
+        //
+        // We plug in the second part of the pipeline, including the deinterleave element,
+        // once the media starts being decoded.
         decodebin.connect_pad_added(move |_, src_pad| {
-            // We only want a sink per audio stream.
+            // Ignore any additional source pads just in case.
             if src_pad.is_linked() {
```
sdroege: This is never going to be true. What you want to check here is if linking to the next sinkpad fails because the pad is already linked

backends/gstreamer/src/audio_decoder.rs
```diff
@@ -97,84 +115,164 @@ impl AudioDecoder for GStreamerAudioDecoder {
...
+                let filter = gst::ElementFactory::make("capsfilter", None).ok_or(())?;
+                let deinterleave =
+                    gst::ElementFactory::make("deinterleave", Some("deinterleave")).ok_or(())?;
+
+                deinterleave
+                    .set_property("keep-positions", &true.to_value())
                     .map_err(|_| ())?;
+                let pipeline_ = pipeline.clone();
+                let callbacks_ = callbacks.clone();
+                deinterleave.connect_pad_added(move |_, src_pad| {
```
sdroege: You create a reference cycle here: pipeline -> deinterleave -> pipeline.

Best to use weak references (`pipeline.downgrade()` and later `pipeline.upgrade()`) for passing into the closure. Same bug elsewhere probably

audio/src/decoder.rs
```diff
@@ -44,13 +54,14 @@ unsafe impl Send for AudioDecoderCallbacks {}
 unsafe impl Sync for AudioDecoderCallbacks {}
 
 pub struct AudioDecoderCallbacksBuilder {
-    eos: Option<Box<FnBox(u32) + Send + 'static>>,
+    eos: Option<Box<FnBox() + Send + 'static>>,
```
Manishearth: can we still pass down channels here so we don't have to use shared mutable state on the DOM side to deal with this?

Manishearth:
I don't understand most of this, I'll defer to @sdroege's review

audio/src/decoder.rs
ferjm: We still need to use shared mutable state to [build](https://github.com/servo/servo/blob/master/components/script/dom/baseaudiocontext.rs#L447) the final decoded audio from the progress callbacks and to [create](https://github.com/servo/servo/blob/master/components/script/dom/baseaudiocontext.rs#L412) the final `AudioBuffer` content when receiving the `eos` callback. I think we can infer the number of channels from the final decoded audio array, so we don't need to send the channel count param through the `eos` callback.

sdroege:
Looks good but you probably want to check for similar circular references in other closures

PR #122, opened 2018-09-10
Add gstreamer deps to travis

ferjm:
r? @Manishearth


Manishearth:
Added the missing target keys, r=me if travis passes

jdm:
```
     Running `target/debug/params_connect2`
thread 'AudioRenderThread' panicked at 'index 128 out of range for slice of length 0', libcore/slice/mod.rs:1932:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', libcore/result.rs:979:5
fatal runtime error: failed to initiate panic, error 5
xargs: cargo: terminated by signal 6
```

ferjm:
> thread 'AudioRenderThread' panicked at 'index 128 out of range for slice of length 0', libcore/slice/mod.rs:1932:5

Hmm... I am not able to reproduce this panic locally :\

Manishearth:
Same, can't repro locally. This is weird.

Maybe it's an intermittent? Can't see why but that would explain why params_connect wasn't affected (it does basically the same thing)

philn:
Might be worth adding the `GST_DEBUG=3` env var as well.

ferjm:
> Maybe it's an intermittent? Can't see why but that would explain why params_connect wasn't affected (it does basically the same thing)

They indeed look intermittents :( `params_connect` has also failed on the last run.

ferjm:
@Manishearth could we merge this PR and file a follow up issue to investigate the intermittents? This way we could at least see if a PR breaks the build.

Manishearth:
Uh, sure, my concern is that this prevents us from updating media in servo
until we get CI to support Android gstreamer.

On Sep 18, 2018 7:01 PM, "Fernando Jiménez Moreno" <notifications@github.com>
wrote:

@Manishearth <https://github.com/Manishearth> could we merge this PR and
file a follow up issue to investigate the intermittents? This way we could
at least see if a PR breaks the build.

—
You are receiving this because you were mentioned.
<https://github.com/servo/media/pull/122#issuecomment-422394175>, or mute
the thread

ferjm:
I don't see how this relates to Android. Are you referring to #129 maybe :)?

Manishearth:
Derp.

Yeah please merge this. Perhaps whitelist the intermittent tests with a
`grep -V`

PR #123, opened 2018-09-11
 Expose getter for AudioBuffer mutable channel data 

ferjm:
We need this to fix an [AudioBuffer WPT crash](https://github.com/servo/servo/blob/master/tests/wpt/metadata/webaudio/the-audio-api/the-audiobuffer-interface/audiobuffer-copy-channel.html.ini)


ferjm:
The Cargo.lock update was missing from #121

PR #131, opened 2018-09-18
Implement BiquadFilterNode

Manishearth:
This contains all the code BiquadFilterNode should need, but I haven't had a chance to test it yet.

However, there are lots of equations here, so I thought I'd post it early to get a head start on review.

Note that for the actual processing code I'm using the formula from https://github.com/WebAudio/web-audio-api/pull/1758 .

r? @ferjm


Manishearth:
Added a test; I verified it works by aurally comparing it with Firefox.

r? @ferjm

PR #137, opened 2018-09-27
Make sure that we can run player code in Travis

ferjm:


ceyusa:
As is a snapshot of master, I would stamp the generation date in the tarball name.

ceyusa:
> As is a snapshot of master, I would stamp the generation date in the tarball name.

I answer myself: since we rollbacked to release 1.14, there's no need to name the tarball with a date.

r+

Issue #125, opened 2018-09-11
Report missing decoders and other audio decoder errors


ferjm:
We should be listening for pipeline bus error messages and firing the audio decoder error callback when needed.

```
philn> ferjm: there you go :)
12:27:54 travis needs ogg/vorbis libs
12:28:21 <ferjm> Missing decoder: Ogg (audio/ogg)
12:28:33 cool
12:30:18 we shouldn't be freezing there though :\
12:33:03 philn: I guess decodebin is throwing errors for missing decoders, right?
12:34:30 ⇐ BK1603 quit (uid316735@moz-jgv72q.stonehaven.irccloud.com) Quit: Connection closed for inactivity
12:36:35 <philn> yeah
12:38:58 you should be able to reproduce the error locally by moving out libgstogg.so
12:39:16 <@Manishearth> philn: thanks :)
12:39:36 <ferjm> philn: ok, cool. Are these errors sent through as a pipeline bus message or is there an error callback I can handle in decodebin?
12:39:59 <@Manishearth> ferjm: in the example we wait for it to respond, we should .unwrap() the .recv() call
12:40:02 that's why it hangs
12:40:03 <philn> if you use gst-player you can connect to the error signal iirc
12:40:08 <@Manishearth> we get stuck on .recv()
12:40:26 <philn> otherwise yes, an error message is sent over the bus
12:41:10 <@Manishearth> philn: its a bug in the test's error handling
12:41:16 → rohitpaulk23833258 joined (loungeuser@moz-emn.5fb.201.103.IP)
12:41:22 <philn> ah, ok!
12:42:15 <ferjm> philn: ok, we are not using gst-player in this case
12:42:43 Manishearth:  https://github.com/servo/media/blob/master/examples/audio_decoder.rs#L50
12:43:17 <@Manishearth> ferjm: oh hm
12:43:18 <ferjm> we should be firing the error callback for the missing decoder and send a message through the channel there https://github.com/servo/media/blob/master/examples/audio_decoder.rs#L34
12:43:23 <@Manishearth> yeah
12:43:34 it's an eror handler so it doesn't panic, my bad
12:43:34 <ferjm> but we need to get the error from the pipeline bus first 
```

sdroege:
What kind of errors would you like the player to export? Currently the error enum there is just a stub. Having separate errors for missing decoders, file not found, etc. would definitely be something I'd like to have in there but I didn't get to design anything yet. Errors are hard :)

[...]

PR #138, opened 2018-09-27
Add proper error handling, make sure sink cannot panic

Manishearth:
Not sure if the decoder side needs this

r? @ferjm

### DIFF
--- a/backends/gstreamer/src/audio_sink.rs
+++ b/backends/gstreamer/src/audio_sink.rs
@@ -23,15 +23,27 @@ pub struct GStreamerAudioSink {
     sample_offset: Cell<u64>,
 }
 
+#[derive(Debug)]
+pub enum BackendError {
+    Gstreamer(gst::Error),
+    Flow(gst::FlowError),
+    ElementCreationFailed(&'static str),
+    AudioInfoFailed,
+    PipelineFailed(&'static str),
+    StateChangeFailed,
+}
+
+
 impl GStreamerAudioSink {
-    pub fn new() -> Result<Self, ()> {
+    pub fn new() -> Result<Self, BackendError> {
         if let Some(category) = gst::DebugCategory::get("openslessink") {
             category.set_threshold(gst::DebugLevel::Trace);
         }
-        gst::init().map_err(|_| ())?;
+        gst::init().map_err(BackendError::Gstreamer)?;
 
-        let appsrc = gst::ElementFactory::make("appsrc", None).ok_or(())?;
-        let appsrc = appsrc.downcast::<AppSrc>().map_err(|_| ())?;
+        let appsrc = gst::ElementFactory::make("appsrc", None)
+                .ok_or(BackendError::ElementCreationFailed("appsrc"))?;
+        let appsrc = appsrc.downcast::<AppSrc>().unwrap();
         Ok(Self {
             pipeline: gst::Pipeline::new(None),
             appsrc: Arc::new(appsrc),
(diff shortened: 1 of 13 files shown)


backends/gstreamer/src/audio_sink.rs
```diff
@@ -23,15 +23,28 @@ pub struct GStreamerAudioSink {
...
             category.set_threshold(gst::DebugLevel::Trace);
         }
-        gst::init().map_err(|_| ())?;
+        gst::init().map_err(BackendError::Gstreamer)?;
 
-        let appsrc = gst::ElementFactory::make("appsrc", None).ok_or(())?;
-        let appsrc = appsrc.downcast::<AppSrc>().map_err(|_| ())?;
+        let appsrc = gst::ElementFactory::make("appsrc", None)
+                .ok_or(BackendError::ElementCreationFailed("appsrc"))?;
+        let appsrc = appsrc.downcast::<AppSrc>().map_err(|_| BackendError::DowncastingFailed)?;
```
sdroege: This is impossible to fail in this case. You create an appsrc so it will be possible to cast it

backends/gstreamer/src/audio_sink.rs
```diff
@@ -43,19 +56,19 @@ impl GStreamerAudioSink {
...
 impl GStreamerAudioSink {
-    fn set_audio_info(&self, sample_rate: f32, channels: u8) -> Result<(), ()> {
+    fn set_audio_info(&self, sample_rate: f32, channels: u8) -> Result<(), BackendError> {
         let audio_info = gst_audio::AudioInfo::new(
             gst_audio::AUDIO_FORMAT_F32,
             sample_rate as u32,
             channels.into(),
         ).build()
-            .ok_or(())?;
+            .ok_or(BackendError::AudioInfoFailed)?;
```
sdroege: This also can't possibly fail. The only way how this can fail if incompatible other settings (e.g. conflicting channel positions) are given

backends/gstreamer/src/audio_sink.rs
```diff
@@ -95,38 +109,42 @@ impl AudioSink for GStreamerAudioSink {
...
+            .ok_or(BackendError::ElementCreationFailed("audioconvert"))?;
+        let sink = gst::ElementFactory::make("autoaudiosink", None)
+            .ok_or(BackendError::ElementCreationFailed("autoaudiosink"))?;
         self.pipeline
             .add_many(&[&appsrc, &resample, &convert, &sink])
-            .map_err(|_| ())?;
-        gst::Element::link_many(&[&appsrc, &resample, &convert, &sink]).map_err(|_| ())?;
+            .map_err(|_| BackendError::PipelineFailed)?;
+        gst::Element::link_many(&[&appsrc, &resample, &convert, &sink])
+            .map_err(|_| BackendError::PipelineFailed)?;
```
sdroege: You maybe want to take the `&'static str` out of the error and store it in `PipelineFailed`

backends/gstreamer/src/audio_sink.rs
```diff
@@ -23,15 +23,28 @@ pub struct GStreamerAudioSink {
...
+        gst::init().map_err(BackendError::Gstreamer)?;
 
-        let appsrc = gst::ElementFactory::make("appsrc", None).ok_or(())?;
-        let appsrc = appsrc.downcast::<AppSrc>().map_err(|_| ())?;
+        let appsrc = gst::ElementFactory::make("appsrc", None)
+                .ok_or(BackendError::ElementCreationFailed("appsrc"))?;
+        let appsrc = appsrc.downcast::<AppSrc>().map_err(|_| BackendError::DowncastingFailed)?;
         Ok(Self {
             pipeline: gst::Pipeline::new(None),
             appsrc: Arc::new(appsrc),
```
sdroege: Why is the appsrc not put into the pipeline here already?

Also generally it seems like for *all* pipelines you're not watching their `gst::Bus` for messages, especially not for error messages. It would be good to add some code for that so that you can handle errors (and later possibly other things) asynchronously while the pipeline is running

ferjm: > Also generally it seems like for all pipelines you're not watching their gst::Bus for messages, especially not for error messages. It would be good to add some code for that so that you can handle errors (and later possibly other things) asynchronously while the pipeline is running

Yeah, we'll need that for #125

backends/gstreamer/src/audio_sink.rs
Manishearth: This feels like it's possible to cause in a refactoring, so I'll leave this error in

player/src/lib.rs
```diff
@@ -2,6 +2,8 @@ extern crate ipc_channel;
 #[macro_use]
 extern crate serde_derive;
 
+extern crate servo_media_audio;
```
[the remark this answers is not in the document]
Manishearth: It's hard because servo-media-audio needs the dummy backend.

However I can make the PlayerBackend trait completely separate. I'll try something.

Manishearth:
Addressed

PR #139, opened 2018-10-01
Handle audio decoder errors

ferjm:


backends/gstreamer/src/audio_decoder.rs
```diff
@@ -296,6 +297,36 @@ impl AudioDecoder for GStreamerAudioDecoder {
         appsrc.set_property_format(gst::Format::Bytes);
         appsrc.set_property_block(true);
 
+        let pipeline_ = pipeline.downgrade();
+        let callbacks_ = callbacks.clone();
+
+        Builder::new()
```
Manishearth: I feel like we should be able to do this in the same thread, but no big deal if we can't.

sdroege: You could integrate it into an existing event loop via e.g. the fd (it becomes readable once you can pop a message), or by using `set_sync_bus_handler()` and from there dispatching the message to your event loop (and returning `Drop`) and not doing anything else

Manishearth: Ah, `set_sync_bus_handler` was what I was looking for. We can have that send an `Error()` message to the event loop

ferjm: Done

PR #145, opened 2018-10-05
Android cleanup

ferjm:
We were previously downloading and setting up the GStreamer dependencies for Android while building the `servo-media-gstreamer` crate. But a user trying to use this crate on Android is also required to fetch the GStreamer dependencies to set up the JNI related stuff. This is the case for Servo since we enabled this dependency for Android. Which means that we are doing this download and setup twice.

This PR moves the responsibility of fetching and setting up the GStreamer dependencies for Android to the user of this crate, so the duplication is avoided.

I added a script to ease the bootstrap process for Android and updated the README accordingly.


etc/android_bootstrap.sh
```diff
@@ -0,0 +1,32 @@
+if [ "$#" -ne 1 ]; then
```
Manishearth: `set -e` please

Issue #147, opened 2018-10-08
AudioScheduledSourceNode handling is buggy and `.start(offset)` doesn't work right


Manishearth:
Currently, `node.start(t)` doesn't seem to work for scheduled source nodes. The following patch fixes it, but overall I'm not sure we're doing the right thing with the two-return-value `should_play_at` function.

cc @ferjm could you look into this and figure out if this is the right fix? I'm using this fix so I can move ahead with my WPT investigation, but we should probably improve this code overall

```diff
diff --git a/audio/src/buffer_source_node.rs b/audio/src/buffer_source_node.rs
index 7847581..68d6ca2 100644
--- a/audio/src/buffer_source_node.rs
+++ b/audio/src/buffer_source_node.rs
@@ -118,10 +118,14 @@ impl AudioNodeEngine for AudioBufferSourceNode {
 
         let len = { self.buffer.as_ref().unwrap().len() as usize };
 
-        if self.playback_offset >= len || self.should_play_at(info.frame) == (false, true) {
+        let should = self.should_play_at(info.frame);
+        if self.playback_offset >= len || should == (false, true) {
             self.maybe_trigger_onended_callback();
             inputs.blocks.push(Default::default());
             return inputs;
+        } else if !should.0 {
+            inputs.blocks.push(Default::default());
+            return inputs;
         }
```

Manishearth:
Oh, another bug here is that this doesn't handle start times within the same buffer.

Manishearth:
I'm going to try and rewrite this stuff today.

PR #148, opened 2018-10-10
Various bugfixes

Manishearth:
Fixes https://github.com/servo/media/issues/147

This fixes almost all the biquad failures, as well as some analyser node and buffer node failures.

More information for why the fix was necessary can be found in the individual commit messages

r? @ferjm


Manishearth:
The test fixes won't show up unless we also fix https://github.com/servo/servo/issues/21887

Manishearth:
Fixed

Issue #151, opened 2018-10-12
Race condition in GStreamer decoder


Manishearth:
The audio_decoder example occasionally panics on [this assertion](https://github.com/servo/media/blob/master/audio/src/buffer_source_node.rs#L201). This happens a lot on CI (which has "1.5 threads" IIRC) but less so locally.

I investigated it a bit (it's pretty easy to reproduce under rr's chaos mode, `rr record -h target/debug/audio_decoder`).

The problem is that gstreamer decodes things on multiple threads (per-channel as far as I can tell), which both call [`new_sample()`](https://github.com/servo/media/blob/b274b9636d4f90e44835735279674c1279e75592/backends/gstreamer/src/audio_decoder.rs#L176). When it reaches the end, [`eos()`](https://github.com/servo/media/blob/b274b9636d4f90e44835735279674c1279e75592/backends/gstreamer/src/audio_decoder.rs#L233) gets called. However, the `new_sample()` from the other thread may not have run yet -- `eos()` isn't guaranteed to run _after_ all of the `new_sample()` calls finish.

One possible fix is to have a second RWlock -- always read-acquired by `new_sample()`. `eos()` write-acquires it and first drains `appsink.pull_sample()` into the buffers before sending the actual EOS signal. This is kinda ugly, I'm hoping there's a way we can ensure eos gets called after everything else, or have a way to wait in eos for everything to finish.

cc @ferjm @sdroege

sdroege:
By using `deinterleave` and `queue`s after it, each of the following `appsink` will get its data from a different thread. That's correct.

The problem here is (if I understand your callbacks correctly) that, while each `appsink` will always have both callbacks called from the same thread (there won't be an `eos` for the same `appsink` during `new-sample`, it will only be triggered *after* the last `new-sample` callback has returned from the very same thread), you have multiple `appsink`. You need to wait for the `eos` callback of *all* of them before considering things to be finished, e.g. by using a simple counter.

I don't know how your exact pipeline looks like, but if the `appsink`s are the only sinks in this pipeline (or you want to wait for all sinks to finish) you could also leave the counting to GStreamer and make use of the `Eos` message from the bus (that message will only appear once *every* sink has received their `Eos` event, the event that triggers the `eos` callback on `appsink). You could wait for it in the sync handler you already have in this code and trigger "global" end-of-stream from there instead of directly from the `appsink`s.

PR #152, opened 2018-10-12
Wait for Eos message on the bus for signalling global end-of-stream

sdroege:
Each appsink will signal eos individually and considering the first
one's eos as the global end-of-stream might cause the pipeline to be
shut down before all other appsinks are done processing their data.

The Eos message on the bus is only sent once *all* sinks are
end-of-stream.

Fixes https://github.com/servo/media/issues/151

----

Untested but should fix the race condition. I've also added the line to send something to the mpsc sender because a few lines below a comment claims that we `// Wait until we get an error or EOS.`, and only the error cases were covered.


Manishearth:
This works locally, I can no longer reproduce the crash.

PR #157, opened 2018-10-18
Player seek

ferjm:


backends/gstreamer/src/player.rs
```diff
@@ -47,6 +47,8 @@ fn metadata_from_media_info(media_info: &PlayerMediaInfo) -> Result<Metadata, ()
         .get_container_format()
         .unwrap_or_else(|| "".to_owned());
 
+    let seekable = media_info.is_seekable();
```
ceyusa: As this flag is highly dependent on if the user set ``set_stream_type()``, but is value is, if the use set Seekable{Fast}, to know if the media container demuxer  is capable of seeking. But in theory we should control the installed demuxers and which are usable by us.

So, I'm not really sure if this metadata is required.

backends/gstreamer/src/player.rs
```diff
@@ -123,6 +123,16 @@ impl PlayerInner {
         self.input_size = size;
     }
 
+    pub fn set_seekable(&mut self, seekable: Seekable) {
+        if let Some(ref appsrc) = self.appsrc {
+            appsrc.set_stream_type(match seekable {
+                Seekable::NonSeekable => gst_app::AppStreamType::Stream,
+                Seekable::Seekable => gst_app::AppStreamType::Seekable,
+                Seekable::SeekableFast => gst_app::AppStreamType::RandomAccess,
```
ceyusa: I'm not sure if SeekableFast makes sense in the Servo's context, since it depends in the ``FetchReponseListener``'s methods.

backends/gstreamer/src/player.rs
```diff
@@ -465,105 +489,43 @@ impl GStreamerPlayer {
     }
 }
 
+macro_rules! inner_player_proxy {
```
ceyusa: fancy!

PR #159, opened 2018-10-24
Workaround set player uri bug

ferjm:
https://github.com/servo/servo/issues/22010#issuecomment-432599657


sdroege:
:+1: but if you can wait until this evening we can put a link to the fix into the comment. Travelling currently

sdroege:
http://cgit.freedesktop.org/gstreamer/gst-plugins-bad/commit/?id=afbbc3a97ec391c6a582f3c746965fdc3eb3e1f3

Fix will be in 1.14.5 / 1.15.1

PR #164, opened 2018-11-21
Implemented constant_source

Avanthikaa:
@arvindmohan9683 @rutvik5 @Priyankha12
Implemented the missing ConstantSource node type. Created DOM interface for the type.
Connecting the constant_source to the underlying backend involves the addition of code in the main servo repository. We will create a pull request for the same once this pull request is approved.


audio/src/constant_source.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub offset: f32,
+}
+
+impl Default for ConstantSourceNodeOptions {
+    fn default() -> Self {
+        ConstantSourceNodeOptions { offset: 1. }
+    }
+}
+
+#[derive(AudioNodeCommon)]
```
ferjm: According to the [spec](https://webaudio.github.io/web-audio-api/#ConstantSourceNode), `ConstantSourceNode` inherits from `AudioScheduledSourceNode`. So we need to derive `AudioScheduledSourceNode` as well:

```rust
#[derive(AudioScheduledSourceNode, AudioNodeCommon)]
```

Deriving `AudioScheduledSourceNode` adds all these [functions](https://github.com/servo/media/blob/master/servo-media-derive/src/lib.rs#L21) (`should_play_at`, `start`, etc.), so you will also need to add the parameters used in them: `start_at`, `stop_at` and `onended_callback`.

I see that you used `GainNode` as an example to develop this node. I would use `OscillatorNode` instead as a better example, as it inherits from `AudioScheduledSourceNode` as well.

audio/src/constant_source.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+impl AudioNodeEngine for ConstantSourceNode {
+    fn node_type(&self) -> AudioNodeType {
+        AudioNodeType::ConstantSourceNode
+    }
+
+    fn process(&mut self, mut inputs: Chunk, info: &BlockInfo) -> Chunk {
+        debug_assert!(inputs.len() == 1);
```
ferjm: I don't think this assert is correct. This is a source node, which means that it will be the one producing chunks of data and it will have no inputs. This should be instead:

```rust
debug_assert!(inputs.len() == 0);
```

audio/src/constant_source.rs
```diff
@@ -0,0 +1,70 @@
...
+        inputs.blocks[0].explicit_silence();
+        if inputs.blocks[0].is_silence() {
```
ferjm: This will always be true given that you call `explicit_silence` in the previous line. But in any case, I don't think you need this check and/or early return here. We will have silence at this point for sure, what we need to do here is to fill that silence with audio data :)

arvindmohan9683:
Hi,
I was wondering what we should call the constantSourceNode's example. For oscillator, the src file was titled 'oscillator_node.rs' and the example as 'oscillator.rs', and this seems like the general convention for the existing files and examples. However, the initial steps asks for the src file for the constant source node to be called 'constant_source.rs', so I was wondering what the example file must be called.

Avanthikaa:
We've made the necessary changes. Please review

ferjm:
>Hi,
I was wondering what we should call the constantSourceNode's example. For oscillator, the src file was titled 'oscillator_node.rs' and the example as 'oscillator.rs', and this seems like the general convention for the existing files and examples. However, the initial steps asks for the src file for the constant source node to be called 'constant_source.rs', so I was wondering what the example file must be called.

Sorry, I missed this comment before.

`constant_source_node.rs` for the implementation and `constant_source.rs` for the example sounds good.

.gitignore
```diff
@@ -2,3 +2,5 @@
 **/*.rs.bk
 **/wget-log
 /gstreamer/
+.DS_Store
+audio/.DS_Store
```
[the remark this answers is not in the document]
Manishearth: You also still need to remove the file from the pull request.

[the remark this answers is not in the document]
Avanthikaa: Which file needs to be removed? Not sure what that means

Avanthikaa:
We have made the changes. Please review.
Thank you

.gitignore
```diff
@@ -2,3 +2,4 @@
 **/*.rs.bk
 **/wget-log
 /gstreamer/
+.DS_Store
```
ferjm: You still have to remove the ` audio/.DS_Store` file (`git rm audio/.DS_Store`)

audio/src/oscillator_node.rs
```diff
@@ -63,6 +63,7 @@ impl OscillatorNode {
             start_at: None,
             stop_at: None,
             onended_callback: None,
+            periodic_wave_options: options.periodic_wave_options,
```
ferjm: It seems that this is unused, so let's remove it for now, please.

examples/channels.rs
```diff
@@ -3,19 +3,26 @@ extern crate servo_media;
...
-    let osc = context.create_node(AudioNodeInit::OscillatorNode(options), Default::default());
+    let mut options = OscillatorNodeOptions::default();
+    let osc = context.create_node(
+        AudioNodeInit::OscillatorNode(options.clone()),
+        Default::default(),
+    );
     options.freq = 213.;
-    let osc2 = context.create_node(AudioNodeInit::OscillatorNode(options), Default::default());
+    let osc2 = context.create_node(
+        AudioNodeInit::OscillatorNode(options.clone()),
```
ferjm: No need to clone `options` here.

examples/channelsum.rs
```diff
@@ -3,21 +3,31 @@ extern crate servo_media;
...
     options.freq = 213.;
-    let osc2 = context.create_node(AudioNodeInit::OscillatorNode(options), Default::default());
+    let osc2 = context.create_node(
+        AudioNodeInit::OscillatorNode(options.clone()),
+        Default::default(),
+    );
     options.freq = 100.;
-    let osc3 = context.create_node(AudioNodeInit::OscillatorNode(options), Default::default());
+    let osc3 = context.create_node(
+        AudioNodeInit::OscillatorNode(options.clone()),
```
ferjm: No need to clone `options` here.

examples/constant_source.rs
```diff
@@ -0,0 +1,96 @@
...
+use servo_media::ServoMedia;
+use std::sync::Arc;
+use std::{thread, time};
+
+fn run_example(servo_media: Arc<ServoMedia>) {
+    let context = servo_media.create_audio_context(Default::default());
+    let dest = context.dest_node();
+
+    let mut cs_options = ConstantSourceNodeOptions::default();
+    cs_options.offset = 0.;
```
ferjm: The default `offset` is already 0, so no need to set it again here.

ferjm: In any case, we can get rid of `cs_options` and do:

```rust
let cs = context.create_node(
    AudioNodeInit::ConstantSourceNode(ConstantSourceNodeOptions::default()),
    Default::default(),
);
```

Avanthikaa:
Made the required changes. Please verify

ferjm:
Thanks!

PR #167, opened 2018-11-29
Check that we have the required elements to instantiate a gst-player

ferjm:
r? @sdroege

### DIFF
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -234,6 +234,14 @@ impl GStreamerPlayer {
             return Ok(());
         }
 
+        // Check that we actually have the elements that we
+        // need to make this work.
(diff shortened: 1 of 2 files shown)


backends/gstreamer/src/player.rs
```diff
@@ -234,6 +234,14 @@ impl GStreamerPlayer {
             return Ok(());
         }
 
+        // Check that we actually have the elements that we
+        // need to make this work.
+        for element in vec!["playbin", "queue"].iter() {
```
sdroege: Looks good, that should cover the worst setup problems on people's systems at least. It might still not be able to decode a specific stream because of missing codecs, but for that you get a proper error during runtime.

ferjm: Yes, I think we are already reporting decoding issues.

Thanks!

PR #171, opened 2018-12-06
Queue a/v data while player is setting up

ferjm:
r? @jdm


philn:
appsrc uses a queue internally, why not increase its size?

ferjm:
That's a great question :D. I'll revert and use appsrc's internal queue. Thanks!

ferjm:
Actually, we can handle this in a different way. We can make the player signal when it's ready and do not start any resource request until we get that signal, this way we won't need to queue any data.

PR #172, opened 2018-12-07
some cleanups and use videoframe API

ceyusa:
* Some cleanups to avoid multiple versions of some dependencies
* Use VideoFrame from gstreamer-video (adding a new dependency). Also, If I understand correctly, this will save a copy of the frame.

### DIFF
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -5,6 +5,7 @@ use gst;
 use gst_app::{self, AppSrcCallbacks, AppStreamType};
 use gst_player;
 use gst_player::{PlayerMediaInfo, PlayerStreamInfoExt};
+use gst_video::{VideoFrame, VideoInfo};
 use ipc_channel::ipc::IpcSender;
 use servo_media_player::frame::{Frame, FrameRenderer};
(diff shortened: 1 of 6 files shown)


ceyusa:
@ferjm  r?

backends/gstreamer/src/player.rs
```diff
@@ -17,16 +18,19 @@ use std::time;
...
+        .get_caps()
+        .and_then(|caps| VideoInfo::from_caps(caps.as_ref()))
+        .ok_or_else(|| ())?;
+    let frame = VideoFrame::from_buffer_readable(buffer, &info).or_else(|_| Err(()))?;
+    let data = frame.plane_data(0).ok_or_else(|| ())?;
+
+    Ok(Frame::new(
+        info.width() as i32,
+        info.height() as i32,
+        Arc::new(data.to_vec()),
```
sdroege: This still copies like it did before by requiring to have a `Vec<u8>` in the API. Ideally you would change the API so that it takes an `Arc<AsRef<[u8]>>` instead: in that case you could simply pass the return value of `buffer.map_readable()` in there.

My old servo branch had it implemented without copying similar to that.

PR #173, opened 2018-12-13
Allow simple_player example to perform random seeks

ferjm:
We were previously expecting only the two seek requests manually triggered from the tests, but we may receive additional random seek requests from the Player, so lets make sure that we are able to handle any number of seek requests.


ceyusa:
r+

PR #174, opened 2018-12-13
Revert "Queue a/v data while player is setting up"

ferjm:
r? @ceyusa 

As we discussed in person, we are already blocking on the Player setup before allowing to push any data, so there's really no need for a data queue.


ceyusa:
as we talked in the last all-hands

r+

PR #175, opened 2018-12-19
Remove RateChanged player event

ferjm:
There is no way the playback rate can change unless the client changes it through the API, so this event is useless and unfortunately the cause of a deadlock.


ceyusa:
r+

ferjm:
Thanks!

PR #176, opened 2018-12-21
Expose a way to implement a backoff protocol, refactor errors and other minor improvements

ferjm:
This PR exposes the `EnoughData` and `NeedData` Player events that allow consumers to implement a backoff protocol.

I made a couple of significant refactors:

- I moved the list of Player observers and renderers out of `PlayerInner` to avoid deadlocks. For more information, you can read [this IRC conversation](https://mozilla.logbot.info/servo-media/20181219#c15742907-c15743272).
- I had to change the way we report errors in general to allow external consumption.

### DIFF
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -200,6 +200,9 @@ impl PlayerInner {
 
     pub fn push_data(&mut self, data: Vec<u8>) -> Result<(), BackendError> {
         if let Some(ref mut appsrc) = self.appsrc {
+            if appsrc.get_current_level_bytes() + data.len() as u64 >= appsrc.get_max_bytes() {
+                return Err(BackendError::EnoughData);
+            }
             let buffer =
                 gst::Buffer::from_slice(data).ok_or_else(|| BackendError::PlayerPushDataFailed)?;
             if appsrc.push_buffer(buffer) == gst::FlowReturn::Ok {
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -1,4 +1,3 @@
-use super::BackendError;
 use glib;
 use glib::*;
 use gst;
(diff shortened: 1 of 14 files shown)


ceyusa:
what worries me about this change is that we are exposing a gstreamer specific behavior, when the interface is supposed to be generic.

ferjm:
Well, despite the event name choice, I don't think this is a Gstreamer specific behavior. What we are exposing is a way for the Player to control the flow of data that it receives. I expect other backends to have equivalent needs, and if they don't we can still just signal the need of data.

Anyway, I am open to suggestions about how to implement a backoff protocol in a different way :)

backends/gstreamer/src/player.rs
```diff
@@ -488,29 +495,29 @@ impl GStreamerPlayer {
...
+                let inner___ = inner_clone.clone();
                 appsrc.set_callbacks(
                     AppSrcCallbacks::new()
+                        .need_data(move |_, _| {
+                            // We block the caller of the setup method until we get
+                            // the first need-data signal, so we ensure that we
+                            // don't miss any data between the moment the client
+                            // calls setup and the player is actually ready to
+                            // get any data.
+                            if !is_ready_.load(Ordering::Relaxed) {
```
Manishearth: use a Once for this

https://doc.rust-lang.org/nightly/std/sync/struct.Once.html

Manishearth: (this doesn't use Relaxed, though -- any reason why Relaxed is okay in this code? You may get double initialization)

backends/gstreamer/src/player.rs
```diff
@@ -248,8 +214,58 @@ impl PlayerInner {
...
+            renderer.lock().unwrap().render(frame.clone());
+        }
+        Ok(())
+    }
+}
+
 pub struct GStreamerPlayer {
     inner: RefCell<Option<Arc<Mutex<PlayerInner>>>>,
+    observers: Arc<Mutex<PlayerEventObserverList>>,
+    renderers: Arc<Mutex<FrameRendererList>>,
```
Manishearth: do both the lists and the individual renderers need to be behind Arcs and mutexes? I feel like having Vec<FrameRenderer> inside the list may be fine here?

ferjm: Each individual renderer is accessed from the [layout](https://dxr.mozilla.org/servo/source/components/script/dom/htmlmediaelement.rs#1490) and [gstreamer](https://github.com/servo/media/pull/176/files#diff-1d803d24b8ea49dbfeb827b92c7f9c28R261) threads, so I think we need to have them behind Arcs and mutexes.

Manishearth: ah, okay then. r=me

PR #177, opened 2019-01-09
Use a Once per Player init

ferjm:
We need an init flag per Player instance, not a global flag. Regarding [this question](https://github.com/servo/media/pull/176#discussion_r244247665) I believe the usage of `Ordering::Relaxed` is fine here as it is harmless to send the unlock message twice.

### DIFF
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -16,7 +16,6 @@ use std::sync::{Arc, Mutex, Once};
 use std::time;
 use std::u64;
 
-static PLAYER_INIT: Once = Once::new();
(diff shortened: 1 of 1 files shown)


ferjm:
This is causing a bunch of WPTs timeouts https://github.com/servo/servo/pull/22522#issuecomment-452648024

ferjm:
@Manishearth done. Not sure if I can get rid of the `Arc` wrapping the `Once`.

Manishearth:
r+

travis doesn't seem to like the PR

ferjm:
Yeah, Travis is broken since the migration. I don't know exactly why yet.

PR #178, opened 2019-01-09
use ToString trait to get GLib error description

ceyusa:
The current way to get the error description will fail in the next version of glib.


ceyusa:
mmh... the CI is broken...

@ferjm r?

PR #179, opened 2019-01-10
Buffering and media cache

ferjm:


Cargo.toml
```diff
@@ -2,10 +2,12 @@
 members = [
   "audio",
   "backends/gstreamer",
+  "backends/gstreamer/source-element",
+  "backends/gstreamer/source-element-plugin",
```
ceyusa: As a gstreamer plugin can hold a set of different elements, and we could end up adding more custom elements for servo, I wouldn't call source-element-plugin, but servo-plugin or so.

backends/gstreamer/source-element/src/source.rs
```diff
@@ -0,0 +1,174 @@
...
+            match self.appsrc {
+                Some(ref appsrc) => Ok(appsrc.$fn_name($arg1)),
+                None => Err(()),
+            }
+        }
+    )
+}
+
+pub struct ServoSrc {
+    debug_category: gst::DebugCategory,
```
ceyusa: If we don't log out anything, imo there's no need to create a debug category

backends/gstreamer/source-element/src/source.rs
```diff
@@ -0,0 +1,174 @@
...
+            "Servo developers",
+        );
+
+        let caps = gst::Caps::new_simple(
+            "video/x-raw",
+            &[
+                ("format", &"BGRA"),
+                ("pixel-aspect-ratio", &gst::Fraction::from((1, 1))),
+            ],
+        );
```
ceyusa: The source element shold have ANY caps at its source pad, because we don't know what we are receiving from the network, later on, in the pipeline, the media type will be discovered.

backends/gstreamer/source-element/src/source.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    fn get_protocols(&self) -> Vec<String> {
+        vec!["servosrc".into()]
+    }
+}
+
+pub fn register(plugin: &gst::Plugin) -> bool {
+    let type_ = register_type(ServoSrcStatic);
+    gst::Element::register(plugin, "servosrc", 0, type_)
```
ceyusa: I don't know for sure, but I guess we can argue if it's required to register an element that will be used only internally in a project and not exposed to the general gstreamer setup

backends/gstreamer/src/player.rs
```diff
@@ -356,12 +356,12 @@ impl GStreamerPlayer {
         // The estimated version for the fix is 1.14.5 / 1.15.1.
         // https://github.com/servo/servo/issues/22010#issuecomment-432599657
         player
-            .set_property("uri", &Value::from("appsrc://"))
+            .set_property("uri", &Value::from("servosrc://"))
```
ceyusa: Perhaps it should be better to move this property set to servosrc directly.

backends/gstreamer/src/player.rs
```diff
@@ -553,70 +553,30 @@ impl GStreamerPlayer {
...
+                    .dynamic_cast::<ServoSrc>()
+                    .expect("Source element is expected to be a servosrc!");
 
-                appsrc.set_max_bytes(MAX_SRC_QUEUE_SIZE);
-                appsrc.set_property_block(false);
+                servosrc.set_max_bytes(MAX_SRC_QUEUE_SIZE);
+                servosrc.set_property_block(false);
 
-                appsrc.set_property_format(gst::Format::Bytes);
+                servosrc.set_property_format(gst::Format::Bytes);
```
ceyusa: If these properties won't change at all, it would better to place them inside servosrc too.

backends/gstreamer/source-element/src/source.rs
```diff
@@ -123,6 +127,36 @@ mod imp {
...
+                            1, -1, 0
+                        );
+                        query.add_scheduling_modes(&[gst::PadMode::Push]);
+                        println!("QUERY {:?}", query);
+                        return gst::PadProbeReturn::Handled;
+                    }
+                }
+                gst::PadProbeReturn::Ok
+            });
+
```
ceyusa: huh? imo this should be handled by the query() virtual  method in the element, not through a probe in ghost pads... dunno... looks odd

backends/gstreamer/src/player.rs
```diff
@@ -20,6 +20,7 @@ use std::time;
 use std::u64;
 
 const MAX_SRC_QUEUE_SIZE: u64 = 50 * 1024 * 1024; // 50 MB.
+const MAX_BUFFER_SIZE: i32 = 500 * 1024;
```
ceyusa: why this magic number?

backends/gstreamer/source-element/src/source.rs
sdroege: You need to register it so that the URI handler implementation of it is found

Cargo.toml
sdroege: Also you don't really need a plugin at all, it only complicates things if this is only supposed to be used from servo itself and not external applications. All you need to do is to call `gst::Element::register()` at some point, no need for all the plugin stuff.

backends/gstreamer/src/player.rs
```diff
@@ -345,27 +347,39 @@ impl GStreamerPlayer {
...
+            None => {
+                return Err(PlayerError::Backend(
+                    "FlagsClass creation failed".to_owned(),
+                ))
+            }
         };
         pipeline
             .set_property("flags", &flags)
-            .map_err(|e| BackendError::SetPropertyFailed(e.0))?;
+            .map_err(|e| PlayerError::Backend(e.0.to_owned()))?;
```
ceyusa: this should be changed to e.to_string() to avoid api breakage

backends/gstreamer/src/player.rs
```diff
@@ -345,27 +347,39 @@ impl GStreamerPlayer {
...
 
         // Set max size for the player buffer.
         pipeline
             .set_property("buffer-size", &MAX_BUFFER_SIZE)
-            .map_err(|e| BackendError::SetPropertyFailed(e.0))?;
+            .map_err(|e| PlayerError::Backend(e.0.to_owned()))?;
```
ceyusa: ditto

Cargo.toml
sdroege: Ah, you do that in https://github.com/servo/media/pull/179/commits/eaf2cdf864867af9679de57ef315eeff1cb5ea62

ceyusa: Yes, this commit oriented review doesn't work... I should have reviewed the whole diff :D

backends/gstreamer/src/source.rs
```diff
@@ -0,0 +1,236 @@
...
+            // 1. be an http, mms, etc. scheme
+            // 2. report that it is "bandwidth limited".
+            //
+            // 1. is not straightforward because we are using a servosrc scheme for now.
+            // This may change in the future if we end up handling http/https/data
+            // URIs, which is what WebKit does.
+            //
+            // For 2. we need to make servosrc handle the scheduling properties query
+            // to report that it "is bandwidth limited".
+            ghost_pad.add_probe(gst::PadProbeType::QUERY_UPSTREAM, |_, probe_info| {
```
sdroege: I'd recommend overriding the query function of the ghost pad. And for the default handling (for all other queries), you can call `gst::Pad::query_default()`.

backends/gstreamer/src/source.rs
```diff
@@ -0,0 +1,236 @@
...
+        fn get_uri(&self, _element: &gst::URIHandler) -> Option<String> {
+            Some("servosrc://".to_string())
+        }
+
+        fn set_uri(
+            &self,
+            _element: &gst::URIHandler,
+            _uri: Option<String>,
+        ) -> Result<(), glib::Error> {
+            Ok(())
```
sdroege: You probably want to check here if the URI passed in is sensible

Cargo.toml
sdroege: Maybe the commits should be squashed together to allow each of the commits to be meaningful :)

ferjm: Yeah, sorry about that :$

backends/gstreamer/src/player.rs
ferjm: I got it from Gecko. I guess we could check disk space and make this more dynamic, but I'd prefer to do that in a follow up.

backends/gstreamer/src/player.rs
ceyusa: d'oh! that's for player... sorry for the noise

backends/gstreamer/src/source.rs
ferjm: I applied @ceyusa's [suggestion](https://github.com/servo/media/pull/179#discussion_r247875364) so I think this is no longer required.

ceyusa:
r+  :)

PR #180, opened 2019-01-11
Decouple playerwrapper for player test

ceyusa:
In order to configure the test player to use GL rendering, it's required to pass GL parameters from the application to the player, before its configuration. Because of that the player must be instantiated after the UI is launched.

This PR does this.

I think it's better to merge it soon in order to avoid more complex rebases later :)


ceyusa:
@ferjm r+?

Issue #166, opened 2018-11-28
Modify player example to use GStreamerGL


ferjm:
Depends on https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/merge_requests/161

PR #182, opened 2019-01-23
Negotiate  GL textures and render them in player example

ceyusa:
This PR goes above PR https://github.com/servo/media/pull/196

It needs to update the CI GStreamer setup, because it requires  a new version of glib and gstreamer-gl library.

It contains

* Port to webrender 0.58 
* Port to gstreamer git head version (0.13.0) -- this implied a complete rewrite of servosrc
* Add a new trait in Frame to hold a reference to the backend's buffer
* Update glutin version tu current git head
* Add a method to set the GL parameters in the backend's player
* Use glsinkbin in gstreamer player in GL parameters are negotiated
* A lot of changes in example player to render gl textures is --gl argument is passed

Fixes: https://github.com/servo/media/issues/166

### DIFF
--- a/backends/gstreamer/Cargo.toml
+++ b/backends/gstreamer/Cargo.toml
@@ -33,6 +33,10 @@ git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 [dependencies.gstreamer-audio]
 git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
+[dependencies.gstreamer-gl]
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
+features = ["egl"]
+
 [dependencies.gstreamer-player]
 git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
--- a/backends/gstreamer/Cargo.toml
+++ b/backends/gstreamer/Cargo.toml
@@ -33,10 +33,6 @@ git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 [dependencies.gstreamer-audio]
 git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
-[dependencies.gstreamer-gl]
-git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
-features = ["egl"]
-
 [dependencies.gstreamer-player]
 git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
@@ -48,6 +44,12 @@ git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
 [dependencies.gstreamer-video]
 git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
+
+[target.'cfg(not(target_os = "linux"))'.dependencies]
+gstreamer-gl = { git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs" }
+
+[target.'cfg(target_os = "linux")'.dependencies]
+gstreamer-gl = { git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs", features = ["egl"] }
 
 [dependencies.ipc-channel]
 version = "0.11"
--- a/backends/gstreamer/Cargo.toml
+++ b/backends/gstreamer/Cargo.toml
@@ -6,6 +6,7 @@ authors = ["The Servo Project Developers"]
 [build-dependencies]
 regex = "1.0"
 zip = "0.3.1"
+lazy_static = "1.0"
 
 [dependencies]
 boxfnonce = "0.1.0"
@@ -13,44 +14,36 @@ boxfnonce = "0.1.0"
 [dependencies.byte-slice-cast]
 version = "0.2"
 
-[dependencies.glib]
-version = "0.6"
-
 [dependencies.glib-sys]
-version = "0.7"
-
-[dependencies.gobject-subclass]
-version = "0.2.1"
-
-[dependencies.gobject-sys]
-version = "0.7"
-
-[dependencies.gst-plugin]
-version = "0.3.1"
-
-[dependencies.gstreamer]
-version = "0.12"
-
-[dependencies.gstreamer-app]
-version = "0.12"
-
-[dependencies.gstreamer-audio]
-version = "0.12"
-
-[dependencies.gstreamer-player]
-version = "0.12"
+git = "https://github.com/gtk-rs/sys"
 
 [dependencies.gstreamer-sys]
-version = "0.6.1"
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys"
+
+[dependencies.glib]
+git = "https://github.com/gtk-rs/glib"
+
+[dependencies.gstreamer]
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
+features = ["subclassing"]
+
+[dependencies.gstreamer-app]
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
+
+[dependencies.gstreamer-audio]
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
+
+[dependencies.gstreamer-player]
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
 [dependencies.gstreamer-sdp]
-version = "0.12"
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
 [dependencies.gstreamer-webrtc]
-version = "0.12"
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
 [dependencies.gstreamer-video]
-version = "0.12"
+git = "https://gitlab.freedesktop.org/gstreamer/gstreamer-rs"
 
 [dependencies.ipc-channel]
 version = "0.11"
--- a/backends/gstreamer/src/audio_decoder.rs
+++ b/backends/gstreamer/src/audio_decoder.rs
@@ -227,7 +227,7 @@ impl AudioDecoder for GStreamerAudioDecoder {
                                         ));
                                         return gst::FlowReturn::Error;
                                     };
-                                    assert_eq!(audio_info.channels(), 1);
+
                                     let positions = if let Some(positions) = audio_info.positions()
                                     {
                                         positions
--- a/backends/gstreamer/src/audio_decoder.rs
+++ b/backends/gstreamer/src/audio_decoder.rs
@@ -160,7 +160,7 @@ impl AudioDecoder for GStreamerAudioDecoder {
 
                 deinterleave
                     .set_property("keep-positions", &true.to_value())
-                    .map_err(|e| AudioDecoderError::Backend(e.to_string()))?;
+                    .expect("deinterleave doesn't have expected 'keep-positions' property");
                 let pipeline_ = pipeline.downgrade();
                 let callbacks_ = callbacks.clone();
                 deinterleave.connect_pad_added(move |_, src_pad| {
@@ -188,7 +188,7 @@ impl AudioDecoder for GStreamerAudioDecoder {
                         )?;
                         let appsink = sink.clone().dynamic_cast::<gst_app::AppSink>().unwrap();
                         sink.set_property("sync", &false.to_value())
-                            .map_err(|e| AudioDecoderError::Backend(e.to_string()))?;
+                            .expect("appsink doesn't handle expected 'sync' property");
 
                         let callbacks_ = callbacks.clone();
                         appsink.set_callbacks(
@@ -277,9 +277,9 @@ impl AudioDecoder for GStreamerAudioDecoder {
                 let caps = audio_info
                     .to_caps()
                     .ok_or(AudioDecoderError::Backend("AudioInfo failed".to_owned()))?;
-                filter.set_property("caps", &caps.to_value()).map_err(|_| {
-                    AudioDecoderError::Backend("Setting caps property failed".to_owned())
-                })?;
+                filter
+                    .set_property("caps", &caps.to_value())
+                    .expect("capsfilter doesn't have expected 'caps' property");
 
                 let elements = &[&convert, &resample, &filter, &deinterleave];
                 pipeline
--- a/backends/gstreamer/src/audio_decoder.rs
+++ b/backends/gstreamer/src/audio_decoder.rs
@@ -194,49 +194,30 @@ impl AudioDecoder for GStreamerAudioDecoder {
                         appsink.set_callbacks(
                             gst_app::AppSinkCallbacks::new()
                                 .new_sample(move |appsink| {
-                                    let sample = match appsink.pull_sample() {
-                                        None => {
-                                            return gst::FlowReturn::Eos;
-                                        }
-                                        Some(sample) => sample,
-                                    };
-
-                                    let buffer = if let Some(buffer) = sample.get_buffer() {
-                                        buffer
-                                    } else {
+                                    let sample =
+                                        appsink.pull_sample().ok_or(gst::FlowError::Eos)?;
+                                    let buffer = sample.get_buffer().ok_or_else(|| {
                                         callbacks_.error(AudioDecoderError::InvalidSample);
-                                        return gst::FlowReturn::Error;
-                                    };
+                                        gst::FlowError::Error
+                                    })?;
 
-                                    let caps = if let Some(caps) = sample.get_caps() {
-                                        caps
-                                    } else {
-                                        callbacks_.error(AudioDecoderError::Backend(
-                                            "Could not get caps from sample".to_owned(),
-                                        ));
-                                        return gst::FlowReturn::Error;
-                                    };
-
-                                    let audio_info = if let Some(audio_info) =
-                                        gst_audio::AudioInfo::from_caps(&caps)
-                                    {
-                                        audio_info
-                                    } else {
+                                    let audio_info = sample
+                                        .get_caps()
+                                        .and_then(|caps| {
+                                            gst_audio::AudioInfo::from_caps(caps.as_ref())
+                                        })
+                                        .ok_or_else(|| {
+                                            callbacks_.error(AudioDecoderError::Backend(
+                                                "Could not get caps from sample".to_owned(),
+                                            ));
+                                            gst::FlowError::Error
+                                        })?;
+                                    let positions = audio_info.positions().ok_or_else(|| {
                                         callbacks_.error(AudioDecoderError::Backend(
                                             "AudioInfo failed".to_owned(),
                                         ));
-                                        return gst::FlowReturn::Error;
-                                    };
-
-                                    let positions = if let Some(positions) = audio_info.positions()
-                                    {
-                                        positions
-                                    } else {
-                                        callbacks_.error(AudioDecoderError::Backend(
-                                            "AudioInfo failed".to_owned(),
-                                        ));
-                                        return gst::FlowReturn::Error;
-                                    };
+                                        gst::FlowError::Error
+                                    })?;
 
                                     for position in positions.iter() {
                                         let buffer = buffer.clone();
@@ -246,14 +227,14 @@ impl AudioDecoder for GStreamerAudioDecoder {
                                             map
                                         } else {
                                             callbacks_.error(AudioDecoderError::BufferReadFailed);
-                                            return gst::FlowReturn::Error;
+                                            return Err(gst::FlowError::Error);
                                         };
                                         let progress = Box::new(GStreamerAudioDecoderProgress(map));
                                         let channel = position.to_mask() as u32;
                                         callbacks_.progress(progress, channel);
                                     }
 
-                                    gst::FlowReturn::Ok
+                                    Ok(gst::FlowSuccess::Ok)
                                 })
                                 .build(),
                         );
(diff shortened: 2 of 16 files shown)


backends/gstreamer/src/audio_decoder.rs
```diff
@@ -195,49 +195,31 @@ impl AudioDecoder for GStreamerAudioDecoder {
...
-                                        }
-                                        Some(sample) => sample,
-                                    };
-
-                                    let buffer = if let Some(buffer) = sample.get_buffer() {
-                                        buffer
-                                    } else {
+                                    let sample =
+                                        appsink.pull_sample().ok_or(gst::FlowError::Eos)?;
+                                    let buffer = sample.get_buffer().ok_or_else(|| {
```
sdroege: Nice to see that these changes also simplified the code quite a bit here :)

backends/gstreamer/src/source.rs
```diff
@@ -1,28 +1,19 @@
...
+use gst_app;
 use url::Url;
 
 const MAX_SRC_QUEUE_SIZE: u64 = 50 * 1024 * 1024; // 50 MB.
 
+// Implementation sub-module of the GObject
 mod imp {
     use super::*;
-    use glib::prelude::*;
     use gst::ElementExt;
```
sdroege: This line shouldn't be needed, it's already in `super::*`

backends/gstreamer/src/source.rs
```diff
@@ -38,114 +29,186 @@ mod imp {
...
+                    let flags =
+                        gst::SchedulingFlags::SEQUENTIAL | gst::SchedulingFlags::BANDWIDTH_LIMITED;
+                    q.set(flags, 1, -1, 0);
+                    q.add_scheduling_modes(&[gst::PadMode::Push]);
+                    true
+                }
+                _ => false,
+            };
+            let ret = if !ret {
+                use gst::GhostPadExt;
```
sdroege: Same for this

backends/gstreamer/src/source.rs
```diff
@@ -38,114 +29,186 @@ mod imp {
...
+
+                let ghost_pad: gst::GhostPad = pad.clone().downcast().unwrap();
+                let target_pad = ghost_pad.get_target().unwrap();
+                target_pad.query(query)
```
sdroege: Simply doing `pad.query_default(...)` should be equivalent

backends/gstreamer/src/source.rs
```diff
@@ -38,114 +29,186 @@ mod imp {
...
+                    q.set(flags, 1, -1, 0);
+                    q.add_scheduling_modes(&[gst::PadMode::Push]);
+                    true
+                }
+                _ => false,
+            };
+            let ret = if !ret {
+                use gst::GhostPadExt;
+
+                let ghost_pad: gst::GhostPad = pad.clone().downcast().unwrap();
```
sdroege: Instead of this clone dance, you can do `let ghost_pad = pad.downcast_ref::<gst::GhostPad>().unwrap()`

backends/gstreamer/src/source.rs
```diff
@@ -38,114 +29,186 @@ mod imp {
...
+            // to report that it "is bandwidth limited".
+            let ret = match query.view_mut() {
+                gst::QueryView::Scheduling(ref mut q) => {
+                    let flags =
+                        gst::SchedulingFlags::SEQUENTIAL | gst::SchedulingFlags::BANDWIDTH_LIMITED;
+                    q.set(flags, 1, -1, 0);
+                    q.add_scheduling_modes(&[gst::PadMode::Push]);
+                    true
+                }
+                _ => false,
```
sdroege: If you move this crate to the 2018 edition (i.e. enable NLL), you can put the `if` block below directly into the `match` block. I.e.

```rust
match query.view_mut() {
    gst::QueryView::Scheduling(ref mut q) => {
        [...]
        true
     },
    _ => pad.query_default(...),
}
```

ceyusa: the problem there is a second mutable borrow of query

sdroege: Yes, NLL solves that :)

ceyusa: aaaahh

http://smallcultfollowing.com/babysteps/blog/2016/04/27/non-lexical-lifetimes-introduction/ :D

sdroege: That's why I said to update to 2018 edition :)

ceyusa:
Right now, WebRTC GStreamer backend doesn't compile: there are multiple errors of this type

```
error[E0623]: lifetime mismatch                                                                                                                                                                                    
  --> backends/gstreamer/src/media_capture.rs:20:36                                                                                                                                                                
   |                                                                                                                                                                                                               
17 |         builder: gst::caps::Builder,                                                                                                                                                                          
   |                  ------------------ this parameter and the return type are declared with different lifetimes...                                                                                               
18 |     ) -> Option<gst::caps::Builder> {                                                                                                                                                                         
   |          --------------------------                                                                                                                                                                           
19 |         match self {                                                                                                                                                                                          
20 |             Constrain::Value(v) => Some(builder.field(name, v)),                                                                                                                                              
   |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ...but data from `builder` is returned here                                                                                                   
                                                                                                                     
```

backends/gstreamer/src/media_capture.rs
```diff
@@ -10,16 +9,22 @@ pub enum Constrain<T> {
 }
 
 impl Constrain<u64> {
-    fn add_to_caps(self, name: &str, min: u64, max: u64, builder: Builder) -> Option<Builder> {
+    fn add_to_caps(
+        &self,
+        name: &str,
+        min: u64,
+        max: u64,
+        builder: gst::caps::Builder,
```
sdroege: You need to specify the lifetime parameter of the builder here and in the return value. And then it either works or you're violating the lifetime requirements in the body of the function :)

ceyusa: thanks!

ceyusa:
Last Friday I confirmed my suspicion that the media's frame should keep a living reference to the associated GStreamer buffer, otherwise the buffer gets unreffed and the texture turns invalid.

Since media's frame is backend's agnostic, it would be hard to keep a reference to a opaque structure, so I tried out to turn frame into a trait.

But soon I got hit by FrameRenderer trait, which function `render()` receives a Frame, which now turned into a trait, FrameRenderer violates the rules of trait object safety. And I'm kind of trapped there.

ceyusa:
@ferjm @jdm r?

backends/gstreamer/src/player.rs
```diff
@@ -375,23 +375,89 @@ impl GStreamerPlayer {
...
+                                    let context = context.get_mut().unwrap();
+                                    let mut s = context.get_mut_structure();
+                                    s.set_value("context", gl_context.to_send_value());
+                                }
+                                el.set_context(&context);
+                            }
+                        }
+                    }
+                    _ => (),
+                }
```
ferjm: A slightly simpler way of doing the same thing:

```rust
match msg.view() {
    MessageView::NeedContext(ctxt) => {
        if let Some(el) =
            msg.get_src().map(|s| s.downcast::<gst::Element>().unwrap())
        {
            use gst_gl::ContextGLExt;
            
            let context_type = ctxt.get_context_type();
            let mut context = gst::Context::new(context_type, true);
            if context_type == *gst_gl::GL_DISPLAY_CONTEXT_TYPE {
                context.set_gl_display(&gl_display);
            } else if context_type == "gst.gl.app_context" {
                let mut s = context.get_mut().unwrap().get_mut_structure();
                s.set_value("context", gl_context.to_send_value());
            };
            el.set_context(&context);
        }
    }
    _ => (),
}
```

backends/gstreamer/src/player.rs
```diff
@@ -23,28 +24,41 @@ use std::u64;
 const MAX_BUFFER_SIZE: i32 = 500 * 1024 * 1024;
 
 struct GStreamerBuffer {
+    isgl: bool,
```
ferjm: uber-nit: `is_gl` :)

sdroege:
> I'd love to hear @sdroege final feedback as well :)

Won't get to that today but I'll put it on my list.

backends/gstreamer/src/player.rs
ceyusa: The problem there is we shouldn't set a void context (not display nor app)

backends/gstreamer/Cargo.toml
```diff
@@ -6,51 +6,48 @@ authors = ["The Servo Project Developers"]
 [build-dependencies]
 regex = "1.0"
 zip = "0.3.1"
+lazy_static = "1.2.0"
```
sdroege: Why 1.2.0 instead of 1.0?

backends/gstreamer/src/audio_decoder.rs
```diff
@@ -194,49 +194,31 @@ impl AudioDecoder for GStreamerAudioDecoder {
...
+                                    let positions = audio_info.positions().ok_or_else(|| {
                                         callbacks_.error(AudioDecoderError::Backend(
                                             "AudioInfo failed".to_owned(),
                                         ));
-                                        return gst::FlowReturn::Error;
-                                    };
+                                        gst::FlowError::Error
+                                    })?;
 
                                     for position in positions.iter() {
```
sdroege: Above you `assert_eq!(audio_info.channels(), 1);` and here you then iterate over the positions. There can only be one, or what am I missing here? :)

backends/gstreamer/src/media_capture.rs
```diff
@@ -129,8 +131,8 @@ impl GstMediaDevices {
         println!("requesting {:?}", caps);
         let f = self.monitor.add_filter(filter, &caps);
         let devices = self.monitor.get_devices();
-        if f != 0 {
-            self.monitor.remove_filter(f);
+        if f.is_some() {
+            let _ = self.monitor.remove_filter(f.unwrap());
```
sdroege:
```rust
if let Some(f) = f {
    let _ = self.monitor.remove_filter(f);
}
```

backends/gstreamer/src/player.rs
```diff
@@ -21,20 +23,48 @@ use std::u64;
...
 const MAX_BUFFER_SIZE: i32 = 500 * 1024 * 1024;
 
+struct GStreamerBuffer {
+    is_gl: bool,
+    frame: gst_video::VideoFrame<gst_video::video_frame::Readable>,
+}
+
+impl Buffer for GStreamerBuffer {
+    fn to_vec(&self) -> Result<FrameData, ()> {
+        // XXX: we support only one plane on one view
```
sdroege: You might want to ensure that you get BGRA or BGR or similar before the sink then. The packed formats are guaranteed to be in a single plane.

And you probably only support a specific format in the end anyway, probably some kind of RGB format?

backends/gstreamer/src/player.rs
```diff
@@ -21,20 +23,48 @@ use std::u64;
...
+    };
+    let buffer = GStreamerBuffer { is_gl, frame };
 
-    Ok(Frame::new(
+    Frame::new(
         info.width() as i32,
         info.height() as i32,
-        Arc::new(data.to_vec()),
-    ))
+        Box::new(Arc::new(buffer)),
```
sdroege: Why do you box the arc? An arc is already like a fancy box for most purposes

backends/gstreamer/src/player.rs
```diff
@@ -324,32 +364,97 @@ impl GStreamerPlayer {
             observers: Arc::new(Mutex::new(PlayerEventObserverList::new())),
             renderers: Arc::new(Mutex::new(FrameRendererList::new())),
             is_ready: Arc::new(Once::new()),
+            gl_context: RefCell::new(None),
+            gl_display: RefCell::new(None),
         }
     }
 
+    fn set_bus_sync_handler(&self, pipeline: &gst::Element) -> Result<(), PlayerError> {
+        use gst::ElementExt;
```
sdroege: Pattern is usually to `use gst::prelude::*;` at the top and never worry about the extension traits

backends/gstreamer/src/player.rs
```diff
@@ -324,32 +364,97 @@ impl GStreamerPlayer {
...
+        let gl_context = self.gl_context.replace(None).unwrap();
+        let gl_display = self.gl_display.replace(None).unwrap();
+
+        bus.set_sync_handler(move |_, msg| {
+            use gst::MessageView;
+
+            match msg.view() {
+                MessageView::NeedContext(ctxt) => {
+                    if let Some(el) = msg.get_src().map(|s| s.downcast::<gst::Element>().unwrap()) {
+                        use gst_gl::ContextGLExt;
```
sdroege: Also for this

backends/gstreamer/src/player.rs
```diff
@@ -324,32 +364,97 @@ impl GStreamerPlayer {
...
+                .build();
+            appsink.set_caps(&caps);
+
+            glsink
+                .set_property("sink", &appsink)
+                .map_err(|e| PlayerError::Backend(e.to_string()))?;
+
+            pipeline
+                .set_property("video-sink", &glsink.to_value())
+                .map_err(|e| PlayerError::Backend(e.to_string()))?;
```
sdroege: If any of these `set_property()` fail you have a programming error. You probably want to make that an `expect()` instead.

backends/gstreamer/src/player.rs
```diff
@@ -753,4 +854,30 @@ impl Player for GStreamerPlayer {
...
+                if display.is_some() {
+                    let dpy = display.unwrap();
+                    let context = gst_gl::GLContext::new_wrapped(
+                        &dpy,
+                        ctxt,
+                        gst_gl::GLPlatform::EGL,
+                        gst_gl::GLAPI::ANY,
+                    );
+
+                    (dpy.dynamic_cast::<gst_gl::GLDisplay>().ok(), context)
```
sdroege: Why does this need a `dynamic_cast()`? `upcast()` should be sufficient

backends/gstreamer/src/source.rs
```diff
@@ -38,113 +28,183 @@ mod imp {
...
+                .expect("Could not create appsrc element");
+
+            let pad_templ = klass.get_pad_template("src").unwrap();
+            let ghost_pad = gst::GhostPad::new_no_target_from_template("src", &pad_templ).unwrap();
+
+            ghost_pad.set_query_function(|pad, parent, query| {
+                ServoSrc::catch_panic_pad_function(
+                    parent,
+                    || false,
+                    |queue, element| queue.query(pad.upcast_ref::<gst::Pad>(), element, query),
```
sdroege: Why is the cast here needed?

backends/gstreamer/src/source.rs
```diff
@@ -38,113 +28,183 @@ mod imp {
...
+    impl ObjectImpl for ServoSrc {
+        glib_object_impl!();
+
+        // Called right after construction of a new instance
+        fn constructed(&self, obj: &glib::Object) {
+            // Call the parent class' ::constructed() implementation first
+            self.parent_constructed(obj);
+
+            let bin = obj.downcast_ref::<gst::Bin>().unwrap();
+            self.add_element(bin, &self.appsrc.clone().upcast())
```
sdroege: Why the clone and upcast? And this should be `bin.add_element(&self.appsrc)`, you want to call the public function and not the virtual method :)

backends/gstreamer/src/webrtc.rs
```diff
@@ -168,10 +165,7 @@ impl GStreamerWebRtcController {
                 None
             })
             .unwrap();
-        self.pipeline
-            .set_state(gst::State::Playing)
-            .into_result()
-            .unwrap();
+        self.pipeline.set_state(gst::State::Playing).unwrap();
```
sdroege: You probably want to handle errors here gracefully instead of panicking :)

examples/player/player_wrapper.rs
```diff
@@ -16,12 +17,37 @@ use std::thread::Builder;
 pub struct PlayerWrapper {
     player: Arc<Mutex<Box<Player>>>,
     shutdown: Arc<AtomicBool>,
+    use_gl: bool,
 }
 
 impl PlayerWrapper {
-    pub fn new(path: &Path) -> Self {
+    fn set_gl_params(player: Arc<Mutex<Box<Player>>>, context: &glutin::Context) -> bool {
```
sdroege: You might want to add some `dyn` here for trait objects. Not necessary but gives a better indication what the thing actually is.

player/src/frame.rs
```diff
@@ -25,7 +40,17 @@ impl Frame {
...
+        match self.data {
+            FrameData::Raw(ref data) => data.clone(),
+            _ => panic!(),
+        }
+    }
+
+    pub fn get_texture_id(&self) -> u32 {
+        match self.data {
+            FrameData::Texture(data) => data,
+            _ => panic!(),
```
sdroege: Maybe give some reason for the panic here and above, or if this is impossible to call maybe `unreachable!()` is a better choice (and it can also take a meaningful string)?

backends/gstreamer/src/source.rs
ceyusa: error[E0308]: mismatched types
   --> backends/gstreamer/src/source.rs:122:50
    |
122 |                     |queue, element| queue.query(pad, element, query),
    |                                                  ^^^ expected struct `gst::Pad`, found struct `gst::GhostPad`
    |
    = note: expected type `&gst::Pad`
               found type `&gst::GhostPad`

sdroege: Ah I see, I'll create an issue for this in GStreamer.

Btw, why do you call it `queue`? That's your source here :)

sdroege: Ah nevermind, this is a problem in your code actually.

```rust
        fn query(
            appsrc.set_property_block(false);
            &self,
            appsrc.set_property_format(gst::Format::Bytes);
            pad: &gst::Pad,
            _element: &gst::Element,
            query: &mut gst::QueryRef,
        ) -> bool {
```

This should be

```rust
        fn query<P: IsA<Pad>>(
            appsrc.set_property_block(false);
            &self,
            appsrc.set_property_format(gst::Format::Bytes);
            pad: &P,
            _element: &gst::Element,
            query: &mut gst::QueryRef,
        ) -> bool {
```

backends/gstreamer/src/source.rs
```diff
@@ -38,113 +28,183 @@ mod imp {
...
+                    let flags =
+                        gst::SchedulingFlags::SEQUENTIAL | gst::SchedulingFlags::BANDWIDTH_LIMITED;
+                    q.set(flags, 1, -1, 0);
+                    q.add_scheduling_modes(&[gst::PadMode::Push]);
+                    true
+                }
+                _ => false,
+            };
+            let ret = if !ret {
+                pad.query_default(None::<&gst::Pad>, query)
```
sdroege: Instead of `None`, this should be `Some(element)` (the third argument to your function here)

backends/gstreamer/src/source.rs
sdroege: You didn't actually change this, did you? :)

backends/gstreamer/src/source.rs
ceyusa: oops... it's my mistake... I guess I solved :)

backends/gstreamer/src/webrtc.rs
ceyusa: I'll leave this comment to @jdm and @Manishearth ;)

jdm: Yeah, dealing with error handling better in webrtc code is a separate issue.

backends/gstreamer/src/source.rs
ceyusa: done!

backends/gstreamer/src/source.rs
```diff
@@ -38,113 +28,180 @@ mod imp {
...
+                gst::QueryView::Scheduling(ref mut q) => {
+                    let flags =
+                        gst::SchedulingFlags::SEQUENTIAL | gst::SchedulingFlags::BANDWIDTH_LIMITED;
+                    q.set(flags, 1, -1, 0);
+                    q.add_scheduling_modes(&[gst::PadMode::Push]);
+                    true
+                }
+                _ => {
+                    if let Some(target) = pad.get_target() {
+                        target.query(query)
```
sdroege: Why not `pad.query_default(Some(element), query)`? This is not the same and you probably want the default behaviour instead of just proxying everything.

examples/player/player_wrapper.rs
```diff
@@ -14,14 +18,45 @@ use std::sync::{Arc, Mutex};
...
-    player: Arc<Mutex<Box<Player>>>,
+    player: Arc<Mutex<Box<dyn Player>>>,
     shutdown: Arc<AtomicBool>,
+    use_gl: bool,
 }
 
 impl PlayerWrapper {
-    pub fn new(path: &Path) -> Self {
+    #[cfg(target_os = "linux")]
+    fn set_gl_params(player: Arc<Mutex<Box<dyn Player>>>, context: &glutin::Context) -> bool {
```
sdroege: Instead of returning `bool`, isn't this a case of failure if you return `false`? If so, some kind of `Result<(), _>` would seem nicer here.

sdroege:
> [Use expect() for programmig errors](/servo/media/pull/182/commits/e60888eaa108fe611a522e18670a2d5ef8663250)

Also typo in this commit message, it's missing a "n" in "programming" :)

backends/gstreamer/src/source.rs
ceyusa: Oops! sorry! You already explained that to me and I stubbornly insist :) fixing it

ceyusa:
I guess all (or most of) the issues were solved :)

sdroege:
> I guess all (or most of) the issues were solved :)

:confetti_ball: Nothing else from my side, I guess :)

PR #184, opened 2019-01-25
Merge in webrtc work

Manishearth:
This adds a WebRtcController abstraction that can be used directly from DOM, along with a Gstreamer backend implementation for it.

We don't yet have a good automated test for this functionality.

r? @jdm @ferjm

### DIFF
--- a/backends/gstreamer/Cargo.toml
+++ b/backends/gstreamer/Cargo.toml
@@ -16,9 +16,6 @@ version = "0.2"
 [dependencies.glib]
 version = "0.6"
 
-[dependencies.failure]
-version = "*"
-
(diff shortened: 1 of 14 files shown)


backends/gstreamer/Cargo.toml
```diff
@@ -28,11 +34,23 @@ version = "0.12"
...
+version = "0.12"
+
+[dependencies.gstreamer-webrtc]
+version = "0.12"
+
 [dependencies.ipc-channel]
 version = "0.11"
 
+[dependencies.lazy_static]
+version = "*"
```
jdm: We should pin this.

PR #186, opened 2019-01-26
Add remote stream callback and simple abstraction for playback.

jdm:
There should be no difference in behaviour as a result of these changes. They fix https://github.com/jdm/media/issues/6.

### DIFF
--- a/backends/gstreamer/src/media_stream.rs
+++ b/backends/gstreamer/src/media_stream.rs
@@ -1,6 +1,6 @@
 use glib::ObjectExt;
-use gst::{self, BinExtManual, ElementExt, GObjectExtManualGst};
-use servo_media_webrtc::MediaStream;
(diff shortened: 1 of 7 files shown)


jdm:
It's not clear to me that WebRtcSignaller is really where on_add_stream belongs, but perhaps we should be looking at renaming it to something like PeerConnection.

backends/gstreamer/src/media_stream.rs
```diff
@@ -1,6 +1,6 @@
 use glib::ObjectExt;
-use gst::{self, BinExtManual, ElementExt, GObjectExtManualGst};
-use servo_media_webrtc::MediaStream;
+use gst::{self, BinExt, BinExtManual, ElementExt, GObjectExtManualGst};
```
sdroege: I'd recommend simply importing `glib::prelude::*` and `gst::prelude::*` instead of the traits manually btw

PR #188, opened 2019-01-27
make the build of examples optional

zimio:
The examples inside of the project have more dependencies than the project itself. This creates the issue that by default, if someone tries to build without any of these dependencies, the build will fail.

A better approach would be to make the build of examples optional, because they are not necessary for the library to function. They are just there for pedagogic purposes. Also, not building the examples all the time will speed up build time in downstream projects. 

With this change if you run the following command no examples will be built.

> cargo build 

If you run the following command all examples will be built.

> cargo build --all

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,4 +9,12 @@ members = [
   "servo-media-derive",
   "webrtc"
 ]
+default-members = [
+  "audio",
+  "backends/gstreamer",
+  "player",
+  "servo-media",
+  "servo-media-derive",
+  "webrtc"
+]
 license = "MPL-2.0"
--- a/README.md
+++ b/README.md
@@ -86,6 +86,10 @@ For macOS, Windows, and Linux, simply run:
 ```bash
 cargo build
 ```
+To also build the xamples:
+```bash
+cargo build --all
+```
 For Android, run:
 ```bash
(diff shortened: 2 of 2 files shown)


Manishearth:
You'll also need to make sure Travis runs this.

I'm not sure if we should merge this, I feel like if you need to build a particular crate you should just cd to that folder and do so, that's usually how workspaces work.

zimio:
>You'll also need to make sure Travis runs this.

You are referring to this? https://github.com/servo/media/pull/188/checks?check_run_id=55297147

> I'm not sure if we should merge this, I feel like if you need to build a particular crate you should just cd to that folder and do so, that's usually how workspaces work.

I don't need to build a particular crate. I want to build everything except the examples. I think the reasons I previously stated make it a worthwhile change. Building the examples is not essential and should not be done by default.

Manishearth:
> You are referring to this

Right, the `cargo build` line in `.travis.yml`  would need to have `--all`

> I want to build everything except the examples

For development? In that case you probably should be building the examples too as the API isn't very stable.

Otherwise just build the servo-media crate, that pulls in everything but examples.

zimio:
Ah, ok. I understand now. Then perhaps this PRQ is not necessary.

PR #193, opened 2019-01-30
Fixup dummy backend code

Manishearth:


Manishearth:
merging early so i can land the main PR

cc @jdm

PR #196, opened 2019-02-08
Clean up and misc fixes

ceyusa:
This pull request is a preparation for the video frames rendering as gl textures in the player example:

* isolate the video sinks setup in a function
* add gst logging capabilities to gstreamer player backend, rather than use print macros
* specify a specific commit id to webrender, since we are using git head, but an update will break our example app
* use glib and gstreamer preludes, this is a general clean up recommended by @sdroege 
* the player example eats all the available memory since a great big image leak
* instead of keeping a vector with pushed frames from player example, add a custom queue with only three frames
* as we are using Arc'ed data, there's no need to return a reference


ceyusa:
@ferjm +?

PR #200, opened 2019-02-13
ulong is u32, not u64

Manishearth:


sdroege:
Where does the `ulong` come from exactly? If there's really a C `unsigned long` / `gulong` somewhere here then this change is wrong. `unsigned long` is 64 bit on all 64 bit architectures except for Windows.

sdroege:
Use `libc::c_ulong` if you need the correct type, btw

Manishearth:
Oh, this is JavaScript ulong, which is u32

sdroege:
Oh how confusing. Nevermind then :)

Issue #191, opened 2019-01-29
New sink code isn't working


Manishearth:
When I try this I get a single frozen videosink, and this error in the logs:

```
0:00:00.650705511 29662 0x7f35a8004ad0 WARN                 basesrc gstbasesrc.c:3055:gst_base_src_loop:<nicesrc3> error: Internal data stream error.
0:00:00.650721783 29662 0x7f35a8004ad0 WARN                 basesrc gstbasesrc.c:3055:gst_base_src_loop:<nicesrc3> error: streaming stopped, reason not-linked (-1)
0:00:00.907446112 29662 0x7f356048a6d0 WARN           basetransform gstbasetransform.c:1415:gst_base_transform_reconfigure:<videoconvert1> warning: not negotiated
0:00:00.907457873 29662 0x7f356048a6d0 WARN           basetransform gstbasetransform.c:1415:gst_base_transform_reconfigure:<videoconvert1> warning: not negotiated
0:00:00.956709545 29662 0x7f35bc02ee80 WARN                 basesrc gstbasesrc.c:3055:gst_base_src_loop:<nicesrc0> error: Internal data stream error.
0:00:00.956748859 29662 0x7f35bc02ee80 WARN                 basesrc gstbasesrc.c:3055:gst_base_src_loop:<nicesrc0> error: streaming stopped, reason not-negotiated (-4)
```

the "not linked" error is probably the relevant one, i suspect this cascades and breaks caps negotiation

the following patch which effectively reverts https://github.com/servo/media/pull/186 fixes the problem. I'm not sure what's going on here, they're almost equivalent.

```patch
diff --git a/backends/gstreamer/src/webrtc.rs b/backends/gstreamer/src/webrtc.rs
index eff0af7..e5df76f 100644
--- a/backends/gstreamer/src/webrtc.rs
+++ b/backends/gstreamer/src/webrtc.rs
@@ -268,11 +268,13 @@ fn handle_media_stream(
             let q = gst::ElementFactory::make("queue", None).unwrap();
             let conv = gst::ElementFactory::make("audioconvert", None).unwrap();
             let resample = gst::ElementFactory::make("audioresample", None).unwrap();
+            let sink = gst::ElementFactory::make("autoaudiosink", None).unwrap();
 
-            pipe.add_many(&[&q, &conv, &resample])?;
-            gst::Element::link_many(&[&q, &conv, &resample])?;
+            pipe.add_many(&[&q, &conv, &resample, &sink])?;
[...]

PR #208, opened 2019-02-21
Make sink code work again

Manishearth:
fixes https://github.com/servo/media/issues/191

### DIFF
--- a/backends/gstreamer/src/media_capture.rs
+++ b/backends/gstreamer/src/media_capture.rs
@@ -157,8 +157,8 @@ fn create_input_stream(
         .get_track(stream_type == StreamType::Video, constraint_set)
         .map(|track| {
             let f = match stream_type {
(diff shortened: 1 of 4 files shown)


Manishearth:
Ready for review

r? @jdm

Manishearth:
The issues were:

 - we shouldn't be vp8/opus encoding stuff going to the auto sinks
 - pipeline play state in webrtc was incorrect (and later: pipeline play state in my sink code was also broken)
 - decodebin was connected directly to the webrtc element instead of to the dynamically produced pad
 - the vaapi thing is broken, so gstreamer1.0-vaapi shouldn't be installed

Manishearth:
Addressed.

Looks like requesting `sink_1` also works, it probably had a bad interaction with some prior code.

I still have to prerequest pads, though.

PR #209, opened 2019-02-22
Invert dependency graph for backends

### LINKED ISSUE #194: Add a cargo feature to force use of the dummy backend
This would make it possible to test on CI and make it easy to build servo without gstreamer support if desired (such as when porting to a new platform).

jdm:
This change has several benefits:
* it makes the backend selection explicit by the user of this library (which ensures that it's easy to test that the dummy backend compiles)
* it moves the platform-based backend selection into its own optional crate and reduces duplication
* it makes the backends depend on servo-media, rather than the reverse, which improves build times for Servo when the public media APIs are not modified

The downside is there are more trait objects instead of static types. My suspicion is that the performance hit of using virtual methods here is dwarfed by the other work being done, but I have not attempted to measure this at all.

Fixes #194.


jdm:
The changes for Servo to use this new model are quite short: https://github.com/servo/servo/compare/master...jdm:media-invert

audio/src/render_thread.rs
```diff
@@ -150,13 +150,15 @@ impl<S: AudioSink + 'static> AudioRenderThread<S> {
...
     {
         let thread = Self::prepare_thread(make_sink, sender.clone(), sample_rate, graph, options);
         match thread {
             Ok(mut thread) => thread.event_loop(event_queue),
-            Err((graph, e)) => {
-                error!(
+            Err((_graph, e)) => {
+                error!("Could not start audio render thread due to error `{:?}`", e);
+                assert!(false);
+                /*error!(
```
ferjm: Remove the commented code, please.

backends/dummy/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["The Servo Project Developers"]
+
+[dependencies]
+boxfnonce = "0.1.0"
+ipc-channel = "0.11"
+servo-media = { path = "../../servo-media" }
+servo-media-audio = { path = "../../audio" }
+servo-media-player = { path = "../../player" }
+servo-media-streams = { path = "../../streams" }
+servo-media-webrtc = { path = "../../webrtc" }
```
Manishearth: newline at eof

servo-media/src/lib.rs
```diff
@@ -1,152 +1,57 @@
...
-impl WebRtcBackend for DummyBackend {
-    type Controller = DummyWebRtcController;
-    fn construct_webrtc_controller(
-        _: Box<WebRtcSignaller>,
-        _: WebRtcController,
-    ) -> Self::Controller {
-        DummyWebRtcController
-    }
+pub trait Backend: Send + Sync {
+    fn make_player(&self) -> Box<Player>;
```
ferjm: uber-nit: s/make_player/create_player/g as all the other methods are `create_*`.

servo-media/Cargo.toml
```diff
@@ -18,5 +18,8 @@ path = "../streams"
 [dependencies.servo-media-webrtc]
 path = "../webrtc"
 
-[target.'cfg(any(all(target_os = "android", target_arch = "arm"), target_arch = "x86_64"))'.dependencies.servo-media-gstreamer]
-path = "../backends/gstreamer"
+#[target.'cfg(any(all(target_os = "android", target_arch = "arm"), target_arch = "x86_64"))'.dependencies.servo-media-gstreamer]
```
Manishearth: remove this?

Manishearth:
The reason we initially chose the other model was mostly that it was easier, and we didn't expect other backends. But the dummy backend is turning out to be pretty important, so it's probably good to have it supported better.

I doubt the trait object costs are major here.

PR #212, opened 2019-02-25
Player shutdown

ferjm:
This should allow us to fix issues like https://github.com/servo/servo/issues/22931

### DIFF
--- a/backends/gstreamer/src/player.rs
+++ b/backends/gstreamer/src/player.rs
@@ -313,6 +313,10 @@ impl PlayerEventObserverList {
             observer.send(event.clone()).unwrap();
         }
     }
+
+    fn clear(&mut self) {
+        self.observers.clear();
+    }
 }
 
 struct FrameRendererList {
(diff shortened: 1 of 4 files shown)


backends/gstreamer/src/player.rs
```diff
@@ -873,4 +881,10 @@ impl Player for GStreamerPlayer {
 
         ret
     }
+
+    fn shutdown(&self) -> Result<(), PlayerError> {
+        self.observers.lock().unwrap().clear();
+        self.renderers.lock().unwrap().clear();
+        self.stop()
```
ceyusa: shouldn't stop be before the observers and renders clearing?

ferjm: Stopping will trigger shutdown related messages (i.e. state change message) that observers may not be able to handle anymore. Since this is an active request to shutdown it is safe to assume that the client already knows about the shutdown related notifications and makes the API easier to use (i.e. the client does not need to keep a shutdown flag to discard shutdown messages).

ceyusa: ok!

ceyusa:
r+

PR #213, opened 2019-02-27
Fix sink pad ordering

### LINKED ISSUE #210: Cannot receive webrtc streams from Servo
When testing this please use https://github.com/servo/media/pull/208 . The bug is present without those changes, but is harder to test since the sink code is broken too.

For testing we're using the signalling server in https://github.com/jdm/gstwebrtc-demos/tree/nossl : `python -m SimpleHTTPServer` in the `js` folder, and `./simple-server.py` in the `signalling` folder. You'll need to remove the data channel code from webrtc.js

Currently we have three webrtc clients:

 - Firefox or any other browser, using its own webrtc implementation, pointed at `localhost:8000`
 - Servo, using servo-medias webrtc implementation, pointed at `localhost:8000` (with `dom.webrtc.enabled` set in prefs.json)
 - The simple_webrtc client, which emulates the javascript running on `localhost:8000`, and can be run against itself (`cargo ex simple_webrtc 8443` and then `cargo ex simple_webrtc 8443 <peer id>`)

With https://github.com/servo/media/pull/208 applied, running `simple_webrtc` against `simple_webrtc` works fine, as does running Firefox against `simple_webrtc`.

Servo does not. The servo side does not yet implement _playing_ streams, so we can expect that to not work, however it does support creating streams and sending them over webrtc, using the same servo-media code that simple_webrtc uses. When you run servo against simple_webrtc:

 - the negotiation runs fine
 - servo does indeed see the streams from simple_webrtc (as a pad-added signal)
 - simple_webrtc does _not_ receive the pad-added signal for servo's streams

Manishearth:
Fixes #210

Based on #208

This implements a version of the fix detailed in https://github.com/servo/media/issues/210#issuecomment-467365523

Instead of unlinking/relinking, we simply keep new streams unlinked until we decide to trigger an offer, at which point we link them to new request pads. The answerer side keeps them unlinked until it gets a remote offer, at which point it also requests pads and links them to the correctly numbered request pads.

A missing piece was that the payload number also has to match, so we keep track of that too.

r? @jdm @slomo (last two commits only)

### DIFF
--- a/backends/gstreamer/src/webrtc.rs
+++ b/backends/gstreamer/src/webrtc.rs
@@ -11,18 +11,57 @@ use servo_media_webrtc::WebRtcController as WebRtcThread;
 use servo_media_webrtc::*;
 use servo_media_streams::MediaStream;
 use std::sync::{Arc, Mutex};
+use std::{cmp, mem};
 
 // TODO:
 // - add a proper error enum
 // - figure out purpose of glib loop
 
+#[derive(Debug, Clone)]
+pub struct MLineInfo {
+    /// The caps for the given m-line
+    caps: gst::Caps,
+    /// Whether or not this sink pad has already been connected
+    is_used: bool,
+    /// The payload value of the given m-line
+    payload: i32,
+}
+
 pub struct GStreamerWebRtcController {
     webrtc: Option<gst::Element>,
     pipeline: gst::Pipeline,
-    has_streams: bool,
+    /// We can't trigger a negotiation-needed event until we have streams, or otherwise
+    /// a createOffer() call will lead to bad SDP. Instead, we delay negotiation.
     delayed_negotiation: bool,
+    /// A handle to the event loop abstraction surrounding the webrtc implementations,
+    /// which lets gstreamer callbacks send events back to the event loop to run on this object
     thread: WebRtcThread,
     signaller: Box<WebRtcSignaller>,
+    /// All the streams that are actually connected to the webrtcbin (i.e., their presence has already
+    /// been negotiated)
+    streams: Vec<Box<MediaStream>>,
+    /// Disconnected streams that are waiting to be linked. Streams are
+    /// only linked when:
+    ///
+    /// - An offer is made (all pending streams are flushed)
+    /// - An offer is received (all matching pending streams are flushed)
+    /// - A stream is added when there is a so-far-disconnected remote-m-line
+    ///
+    /// In other words, these are all yet to be negotiated
+    ///
+    /// See link_stream
+    pending_streams: Vec<Box<MediaStream>>,
+    /// Each new webrtc stream should have a new payload/pt value, starting at 96
+    ///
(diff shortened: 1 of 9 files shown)


Manishearth:
I'm done for the day, but I do need to write some more comments for this. Will do tomorrow.

backends/gstreamer/src/webrtc.rs
```diff
@@ -146,14 +177,128 @@ impl GStreamerWebRtcController {
...
+                for _ in self.request_pad_counter..=idx {
+                    // webrtcbin needs you to request pads (or use element.link(webrtcbin))
+                    // however, it also wants them to be connected in the correct order.
+                    //
+                    // Here, we make sure all the numbered sink pads have been created beforehand, up to
+                    // and including the one we need here.
+                    //
+                    // An alternate fix is to sort pending_streams according to the m-line index
+                    // and just do it in order. This also seems brittle.
+                    webrtc.get_request_pad("sink_%u").unwrap();
```
Manishearth: I feel like there's an underlying bug here, this should be unnecessary

cc @slomo

webrtc/src/thread.rs
```diff
@@ -90,6 +90,7 @@ pub enum InternalEvent {
     OnNegotiationNeeded,
     OnIceCandidate(IceCandidate),
     OnAddStream(Box<MediaStream>),
+    DescriptionAdded(SendBoxFnOnce<'static, ()>, bool, SdpType),
```
jdm: Maybe use an enum instead of a bool here?

backends/gstreamer/src/webrtc.rs
```diff
@@ -146,14 +194,128 @@ impl GStreamerWebRtcController {
...
+        let caps = stream.caps();
+        let webrtc = self.webrtc.as_ref().unwrap();
+        let idx = self.remote_mline_info.iter().enumerate()
+            .filter(|(_, x)| !x.is_used)
+            .find(|(_, x)| x.caps.can_intersect(&caps))
+            .map(|x| x.0);
+        let element = stream.src_element();
+
+        if let Some(idx) = idx {
+            if idx >= self.request_pad_counter {
```
jdm: Is there any risk that we could call link_stream with !request_new_pads but end up running this code that requests pads? Would that be bad? I find the logic here a little bit obtuse.

Manishearth: Yes, this code requesting new pads is fine, `request_new_pads` is about requesting new pads when there is no remote m-line: in that case we normally want to just buffer it up for later, except for when we're making an offer, at which point we want to flush all buffered streams.

The behavior of this function with `!request_new_pads` is "request pads only if they have been set up by the remote already", which is too long to fit in a variable. I can document this more explicitly.

slomo:
@Manishearth You are mentioning the wrong guy in your PR.

Manishearth:
Oh, sorry about that!

Issue #206, opened 2019-02-19
Handle webrtc errors better


jdm:
There are a _lot_ of unwraps in the webrtc gstreamer backend because it was the fastest way to experiment. We should address that.

Manishearth:
I'm going to convert a bunch of these to expects(), and add a Webrtcerror type for the rest.

PR #215, opened 2019-03-05
Add proper Webrtc error type

Manishearth:
fixes https://github.com/servo/media/issues/206


Manishearth:
I'm not sure exactly which things should use `?` and which should be `expect()`.

The errors are just printed out and not reported to the client, there _is_ a burden of ensuring things happen in the correct order on the client side, however due to the messaging model there's no nonblocking way to let this library handle errors and report back.

PR #219, opened 2019-03-11
Workaround missing cargo `target-specific features` feature for gstreamer-gl

ferjm:


backends/gstreamer/src/player.rs
```diff
@@ -847,26 +849,28 @@ impl Player for GStreamerPlayer {
         self.renderers.lock().unwrap().register(renderer);
     }
 
+    #[cfg(not(target_os = "linux"))]
+    fn set_gl_params(&self, _: GlContext, _: usize) -> Result<(), ()> {
+        unimplemented!()
```
jdm: Let's have this return an Err instead, perhaps?

PR #223, opened 2019-03-25
Temporarily disable GL functionality

ferjm:
This should unblock updating servo-media in Servo while we work on a different workaround for the [existing build issues](https://github.com/servo/servo/pull/22944#issuecomment-468827665).


ceyusa:
ok :(

r+

PR #226, opened 2019-03-29
Avoid race between set-remote-description and link_streams

Manishearth:
It is possible for link_streams to be called when our code has received a remote offer (and thus `remote_mline_info` is set), but before `set-remote-description` has completely finished executing. This can lead to `link_streams` _thinking_ that the webrtcbin is ready to create new pads, while the internal transceivers haven't yet been created. Thus even though we link streams to the appropriate pads, the transceivers are created in the wrong order, and some of the gstreamer code looks at transceiver creation order instead of mline info (this is a bug), breaking things.

fixes ##225

r? @jdm


jdm:
@bors-servo r+
Thanks for investigating!

PR #227, opened 2019-03-29
Play media streams

ferjm:


backends/gstreamer/src/media_stream_source.rs
```diff
@@ -0,0 +1,250 @@
...
+        video_proxysrc: gst::Element,
+        audio_srcpad: gst::GhostPad,
+        video_srcpad: gst::GhostPad,
+    }
+
+    impl ServoMediaStreamSrc {
+        pub fn set_audio_stream(&self, stream: &GStreamerMediaStream) {
+            debug_assert!(stream.type_() == StreamType::Audio);
+            gst_log!(self.cat, "Setting audio stream");
+            // XXX Do proxysrc stuff.
```
Manishearth: What _should_ work is simply using [`attach_to_pipeline`](https://github.com/servo/media/blob/62d748a799e1df240b661d37b758ced9830ef999/backends/gstreamer/src/media_stream.rs#L94-L103) and then linking the last element to your player element [like so](https://github.com/servo/media/blob/62d748a799e1df240b661d37b758ced9830ef999/backends/gstreamer/src/media_stream.rs#L223)

backends/gstreamer/src/media_stream_source.rs
```diff
@@ -0,0 +1,211 @@
...
+        proxysrc: gst::Element,
+        srcpad: gst::GhostPad,
+    }
+
+    impl ServoMediaStreamSrc {
+        pub fn set_stream(&self, stream: &mut GStreamerMediaStream) {
+            gst_log!(self.cat, "Setting stream");
+
+            // Append a proxysink to the media stream pipeline.
+            let pipeline = stream.pipeline_or_new();
```
Manishearth: Can we just use `attach_to_pipeline` here? Is there a reason there needs to be a proxysrc/sink? The GstreamerMediaStream may already arise from a proxysrc element.

Manishearth:
I can verify that the audio reaches the other end of the proxysink. Unsure where it's getting dropped.

backends/gstreamer/src/media_stream_source.rs
ferjm: What's the benefit of using `attach_to_pipeline` over the current approach?

In any case, I am not sure if this is possible. The GstPlayer [pipeline's](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-bad-libs/html/GstPlayer.html#gst-player-get-pipeline) (playbin) source is [read-only](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-base-plugins/html/gst-plugins-base-plugins-playbin.html#GstPlayBin--source), so I am not sure if I can modify the pipeline directly. I can try if you think it is a better option though.

ferjm:
I can hear sound now with the current patch.

This is not finished yet though. We need to share the mediastream with multiple consumers at the same time (i.e. player, webrtc), so I need to make the mediastream pipeline include a `tee` so we can have one src pad per consumer.

backends/gstreamer/src/player.rs
```diff
@@ -290,6 +292,30 @@ impl PlayerInner {
...
+            if let PlayerSource::Stream(source) = source {
+                if let Some(mut stream) = stream.as_mut_any().downcast_mut::<GStreamerMediaStream>()
+                {
+                    let playbin = self
+                        .player
+                        .get_pipeline()
+                        .dynamic_cast::<gst::Pipeline>()
+                        .unwrap();
+                    let clock = gst::SystemClock::obtain();
+                    playbin.set_base_time(clock.get_time());
```
philn: This base time might not be exactly the same as the one set on the mediastream pipeline. You'd need to share the value, somehow.

backends/gstreamer/src/media_stream_source.rs
Manishearth: > What's the benefit of using attach_to_pipeline over the current approach?

It avoids an extra proxy.

You don't add it to the playbin's pipeline, you add it to `ServoMediaStreamSrc::constructed`. May need a new version of `attach_to_pipeline` for `attach_to_bin`.

Right now the code adds a proxysrc to `ServoMediaStreamSrc`s bin, instead we should attach the entire `MediaStream` pipeline to it.

This doesn't have to be done in this PR, and I can try to do it later if you want.

Manishearth: I guess setting up the ghost pads is trickier if you want to do this, since you don't have the streams during construction. Hmm.

backends/gstreamer/src/player.rs
Manishearth: I think we'll probably try to set basetime zero everywhere to be consistent?

Manishearth:
Is it possible to have multiple `proxysrc`s associated with a single proxysink? If so we won't need the tee element, instead we can unify all mediastreams as a single proxysrc element.

(This is particularly nice for the webrtc case, where we have a proxysrc element already)

Of course, for non-gst streams we will have to use an appsrc-proxysrc pipeline, but we don't have any of those yet.

ferjm:
> Is it possible to have multiple proxysrcs associated with a single proxysink? If so we won't need the tee element, instead we can unify all mediastreams as a single proxysrc element.

That would be nice, and I also thought about this, but @philn told me that proxysrcs have a 1:1 relation with proxysinks, so AFAICT we still need the tee if we want to share streams between different pipelines.

backends/gstreamer/src/media_stream_source.rs
ferjm: Ok, I see.

I think that should be fine as long as we set the ghost pad before setting the player's uri.

Manishearth: Oh, as in we just don't connect the ghost pad to anything? That could work, I'm not sure if ghost pads can do that.

ferjm: We need to connect the ghost pad to something and I thought that that something could be a `fakesrc` or such. But it does not seem to work :\

Manishearth: Yeah. ANyway, we don't have to solve this yet.

ferjm:
Unfortunately I wasn't able to successfully run the webrtc example yet, so I'm going to leave the part to allow multiple consumers of a media stream for a follow-up.

r? @Manishearth

backends/gstreamer/media_stream.rs
```diff
@@ -107,9 +102,10 @@ impl GStreamerMediaStream {
             pipeline.clone()
         } else {
             let pipeline = gst::Pipeline::new("gstreamermediastream fresh pipeline");
+            let clock = gst::SystemClock::obtain();
             pipeline.set_start_time(gst::ClockTime::none());
-            pipeline.set_base_time(gst::ClockTime::from_nseconds(0));
-            pipeline.use_clock(Some(&gst::SystemClock::obtain()));
+            pipeline.set_base_time(*BACKEND_BASE_TIME);
```
Manishearth: Is this change necessary? As I understand it the base times just need to match, they don't need to be the actual time things started.

To be clear: I don't want you to undo this change, I think it's a good step forward to use some semblance of a real base time, I'm more curious if `BACKEND_BASE_TIME` is really necessary (or if we can get away with using 0 everywhere)

Manishearth:
I'll look into getting this working with webrtc.

Manishearth:
If you want multiple media stream consumers, perhaps try it with video + 2x autovideosink?

backends/gstreamer/media_stream.rs
ferjm: I'll defer the answer to @philn :)

philn: The base time can remain at 0 only if both pipelines start playback at the same time, which is not the case here, AFAIU? This is mentioned in the proxy docs (example): https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gst-plugins-bad-plugins/html/gst-plugins-bad-plugins-proxysrc.html

PR #229, opened 2019-03-31
Flatten directory structures

Manishearth:
The `src/` model is nice for when your repository has a single crate in it, and the root directory gets cluttered with a bunch of random other files making it hard to find your code. However, this repo contains many crates all in their own folders, so there's no need for having `src/` here. This is similar to what Servo does for all of its `components/` crates, files fall directly under `components/foo` and not `components/foo/src`.

Overall this makes this repository nicer to navigate.

r? @ferjm


Manishearth:
This will cause some conflicts in https://github.com/servo/media/pull/227, I'm open to landing this without the last three commits if you don't want to deal with that (git's tooling around rename rebases can be weird)

Manishearth:
We should also probably consistently name the crates, but I'll do that later.

ferjm:
> This will cause some conflicts in #227, I'm open to landing this without the last three commits if you don't want to deal with that (git's tooling around rename rebases can be weird)

I would appreciate that, thanks. I can apply the changes after #227 lands.

Manishearth:
done.
-Manish Goregaokar

ferjm:
Thanks!

@bors-servo r+

Issue #204, opened 2019-02-15
Implement ValueCurveAtTime automation


jdm:
This is the processing algorithm described by https://webaudio.github.io/web-audio-api/#dom-audioparam-setvaluecurveattime. This requires adding a new enum variant to UserAutomationEvent, then adding a new AutomationEvent variant, updating UserAutomationEvent::to_event, and processing the new automation event appropriately in AutomationEvent::run.

PR #230, opened 2019-04-02
 Implemented SetValueCurveAtTime

JHBalaji:
fixes https://github.com/servo/media/issues/204

### DIFF
--- a/audio/param.rs
+++ b/audio/param.rs
@@ -256,25 +256,30 @@ pub enum RampKind {
     Exponential,
 }
 
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, PartialEq, Debug)]
 /// https://webaudio.github.io/web-audio-api/#dfn-automation-event
 pub(crate) enum AutomationEvent {
     SetValue(f32),
     SetValueAtTime(f32, Tick),
     RampToValueAtTime(RampKind, f32, Tick),
     SetTargetAtTime(f32, Tick, /* time constant, units of Tick */ f64),
+    SetValueCurveAtTime(
+        Vec<f32>,
+        /* start time */ Tick,
+        /* duration */ Tick,
+    ),
     CancelAndHoldAtTime(Tick),
     CancelScheduledValues(Tick),
 }
 
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, PartialEq, Debug)]
 /// An AutomationEvent that uses times in s instead of Ticks
 pub enum UserAutomationEvent {
     SetValue(f32),
     SetValueAtTime(f32, /* time */ f64),
     RampToValueAtTime(RampKind, f32, /* time */ f64),
     SetTargetAtTime(f32, f64, /* time constant, units of s */ f64),
-    // SetValueCurveAtTime(Vec<f32>, Tick, /* duration */ Tick)
+    SetValueCurveAtTime(Vec<f32>, /* start time */ f64, /* duration */ f64),
     CancelAndHoldAtTime(f64),
     CancelScheduledValues(f64),
 }
@@ -288,6 +293,13 @@ impl UserAutomationEvent {
             }
             UserAutomationEvent::RampToValueAtTime(kind, val, time) => {
                 AutomationEvent::RampToValueAtTime(kind, val, Tick::from_time(time, rate))
+            }
+            UserAutomationEvent::SetValueCurveAtTime(values, start, duration) => {
+                AutomationEvent::SetValueCurveAtTime(
+                    values,
+                    Tick::from_time(start, rate),
+                    Tick::from_time(duration, rate),
+                )
             }
             UserAutomationEvent::SetTargetAtTime(val, start, tau) => {
                 AutomationEvent::SetTargetAtTime(
@@ -311,6 +323,7 @@ impl AutomationEvent {
     pub fn time(&self) -> Tick {
         match *self {
             AutomationEvent::SetValueAtTime(_, tick) => tick,
+            AutomationEvent::SetValueCurveAtTime(_, start, _) => start,
             AutomationEvent::RampToValueAtTime(_, _, tick) => tick,
             AutomationEvent::SetTargetAtTime(_, start, _) => start,
             AutomationEvent::CancelAndHoldAtTime(t) => t,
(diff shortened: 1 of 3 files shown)


Akhilesh1996:
Hi,

We have finished implementing the SetValueCurveAtTime function ([https://github.com/JHBalaji/media-1/blob/master/audio/param.rs#L418-L436](url)) and written an example ([https://github.com/JHBalaji/media-1/blob/master/examples/set_value_curve.rs](url)) to check the functioning of the same.

examples/set_value_curve.rs
```diff
@@ -0,0 +1,110 @@
...
+use servo_media::ServoMedia;
+use std::sync::Arc;
+use std::{thread, time};
+
+fn run_example(servo_media: Arc<ServoMedia>) {
+    let context = servo_media.create_audio_context(Default::default());
+    let dest = context.dest_node();
+
+    //Initializing the values vector for SetValueCurve function
+    let mut values: Vec<f32> = Vec::new();
```
Manishearth: You can use `let values = vec![0., 0., 0., 0., 1., 1., 1., ....]`

audio/param.rs
```diff
@@ -324,6 +329,7 @@ impl AutomationEvent {
         match *self {
             AutomationEvent::SetValueAtTime(_, tick) => Some(tick),
             AutomationEvent::RampToValueAtTime(_, _, tick) => Some(tick),
+            AutomationEvent::SetValueCurveAtTime(_, _, tick) => Some(tick),
```
Manishearth: This should be `start + duration`

audio/param.rs
```diff
@@ -256,25 +256,26 @@ pub enum RampKind {
...
 
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, PartialEq, Debug)]
 /// https://webaudio.github.io/web-audio-api/#dfn-automation-event
 pub(crate) enum AutomationEvent {
     SetValue(f32),
     SetValueAtTime(f32, Tick),
     RampToValueAtTime(RampKind, f32, Tick),
     SetTargetAtTime(f32, Tick, /* time constant, units of Tick */ f64),
+    SetValueCurveAtTime(Vec<f32>, Tick, /* duration */ Tick),
```
Manishearth: both these variants should have a comment saying that the first `Tick` is the start time

audio/param.rs
```diff
@@ -408,6 +415,25 @@ impl AutomationEvent {
                 *value = val + (event_start_value - val) * exp.exp() as f32;
                 true
             }
+            AutomationEvent::SetValueCurveAtTime(ref values, start, duration) => {
+                let time_diff = ((duration.0 as f32) - (start.0 as f32)) as f32;
```
Manishearth: `duration` is already the time difference here

audio/param.rs
```diff
@@ -408,6 +415,25 @@ impl AutomationEvent {
...
+                let mut progress = ((((current_tick.0 as f32) - (start.0 as f32)) as f32) / time_diff) as f32;
+                if progress < 0.0 {
```
Manishearth: This should never be negative, perhaps we can add a `debug_assert!()` to check that instead of silently setting it to zero?

audio/param.rs
```diff
@@ -408,6 +415,25 @@ impl AutomationEvent {
...
+                    progress = 0.0 as f32;
+                }
+                let n = values.len() as f32;
+	            let k = (((n - 1.) * progress) as f32).floor();
+                let next = k + 1. as f32;
+                let step = time_diff / (n - 1.);
+                if next < n {
+                    let time_k = (k * step) as f32;
```
Manishearth: This is recalculating stuff we already know, an easier way to do this is to calculate `k_float` with `k = k_float.floor()`, and then interpolate between `v[k]` and `v[k+1] with progress `k_float - k`.

audio/param.rs
```diff
@@ -263,7 +263,11 @@ pub(crate) enum AutomationEvent {
     SetValueAtTime(f32, Tick),
     RampToValueAtTime(RampKind, f32, Tick),
     SetTargetAtTime(f32, Tick, /* time constant, units of Tick */ f64),
-    SetValueCurveAtTime(Vec<f32>, Tick, /* duration */ Tick),
+    SetValueCurveAtTime(
+        Vec<f32>,
+        /* start time, units of Tick */ Tick,
+        /* duration, units of Tick */ Tick,
```
Manishearth: (no need to clarify the unit, the type is the unit)

Akhilesh1996: Hi Manish,

Thank you for the comments, I don't quite understand how to implement the k_float stuff. Do you want me to get rid of the step variable and just store (current_tick.0 as f32) - (start.0 as f32) in t_k (which will be progress * duration) and something else in t_k_next? If so, I don't understand how I can calculate the t_k_next value without using the step value.

Can you help me out with this?

Thanks.

Manishearth: You don't need `time_k_next`

You already have `k_float - k`, which is a number 0 to 1 representing where to interpolate.

So you can do something like

```rust
if k + 1 < n {
   let progress = k_float - k
   *value = v[k] * (1 - progress) + v[k + 1] * progress;
} else {
  *value = v[n - 1];
}
```

Akhilesh1996: Hi Manish,

Thank you. I have made the changes and ran rustfmt. Is it fine now?

Akhilesh1996:
Hi Manish, 

Thank you. 

We had first cloned the repo and added commits, after that forked the repo again and pushed changes to it. So there are 6 commits made to the original servo/media in-between the commits made by us. How would you advise us to rebase and squash our commits into a single one?

Thanks
Akhilesh

Manishearth:
In general, instead of merging new commits from master you should try to
rebase.

To fix the current situation, rebase over master first, ensure that that's
working fine (you should have all your commits on top), and then squash
your commits with a second rebase. Make sure your master is updated.

Something like:

```
git fetch origin
git rebase -i origin/master
# check commits are on top
git rebase -i origin/master
# squash
```

you can do both rebases as a single step but it's easier to do it as two so
you can ensure it happened correctly

JHBalaji:
Thanks. Would report back and hopefully in a while.

JHBalaji:
@Manishearth  The merge between the repos and branches we did earlier seems to causing trouble with Squashing. However I have synced with upstream now. Could you please advise on how to squash the commits in this situation?

Manishearth:
You had some merge conflicts that needed to be worked out. I pushed a fix.

In the future, please rebase over master instead of merging in master.

JHBalaji:
Thank you. We would take that note.

PR #232, opened 2019-04-02
Implement can_play_type for ServoMedia

georgeroman:
This is related to https://github.com/servo/servo/issues/22299. I tried to follow WebKit's [implementation](https://github.com/WebKit/webkit/blob/ae5a8bdded113530fe844e01ad9c0e72ce85dbe2/Source/WebCore/platform/graphics/gstreamer/GStreamerCommon.cpp#L304) of a similar function, as suggested by @ferjm in the issue's comments.

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -3,6 +3,7 @@
 
 extern crate boxfnonce;
 extern crate byte_slice_cast;
+extern crate mime;
 
 extern crate glib_sys as glib_ffi;
 extern crate gstreamer_sys as gst_ffi;
@@ -35,13 +36,16 @@ pub mod media_capture;
 pub mod media_stream;
 mod media_stream_source;
 pub mod player;
+mod registry_scanner;
 mod render;
 mod source;
 pub mod webrtc;
 
 use gst::ClockExt;
 use media_stream::GStreamerMediaStream;
-use servo_media::{Backend, BackendInit};
+use mime::Mime;
+use registry_scanner::GSTREAMER_REGISTRY_SCANNER;
+use servo_media::{Backend, BackendInit, SupportsMediaType};
 use servo_media_audio::context::{AudioContext, AudioContextOptions};
 use servo_media_audio::decoder::AudioDecoder;
 use servo_media_audio::sink::AudioSinkError;
@@ -89,6 +93,29 @@ impl Backend for GStreamerBackend {
 
     fn create_videoinput_stream(&self, set: MediaTrackConstraintSet) -> Option<MediaStreamId> {
         media_capture::create_videoinput_stream(set)
+    }
+
+    fn can_play_type(&self, media_type: &str) -> SupportsMediaType {
+        if let Ok(mime) = media_type.parse::<Mime>() {
+            let mime_type = mime.type_().as_str().to_owned() + "/" + mime.subtype().as_str(); 
+            let codecs = match mime.get_param("codecs") {
+                Some(codecs) => {
+                    codecs.as_str().split(',').map(|codec| codec.trim()).collect() 
+                },
+                None => vec![],
+            };
+
+            if GSTREAMER_REGISTRY_SCANNER.is_container_type_supported(&mime_type) {
+                if codecs.is_empty() {
+                    return SupportsMediaType::Maybe;
+                } else if GSTREAMER_REGISTRY_SCANNER.are_all_codecs_supported(&codecs) {
+                    return SupportsMediaType::Probably;
+                } else {
+                    return SupportsMediaType::No;
+                }
+            }
+        }
+        SupportsMediaType::No
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1255,6 +1255,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "mime"
+version = "0.3.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "unicase 2.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "miniz_oxide"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1975,6 +1983,7 @@ dependencies = [
  "gstreamer-webrtc 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ipc-channel 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "mime 0.3.13 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-media 0.1.0",
  "servo-media-audio 0.1.0",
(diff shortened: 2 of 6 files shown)


backends/gstreamer/src/lib.rs
```diff
@@ -80,6 +80,25 @@ impl Backend for GStreamerBackend {
     fn create_videoinput_stream(&self, set: MediaTrackConstraintSet) -> Option<Box<MediaStream>> {
         media_capture::create_videoinput_stream(set).map(|s| Box::new(s) as Box<MediaStream>)
     }
+
+    fn can_play_type(&self, media_type: &str) -> bool {
+        match gst::caps::Caps::from_string(media_type) {
+            Some(caps) => {
+                let factories = gst::ElementFactory::list_get_elements(
+                    gst_ffi::GST_ELEMENT_FACTORY_TYPE_ANY,
```
georgeroman: I was not sure what to pass here. Should this parameter get passed as an argument?

ceyusa:
I would like rather to have a fixed list of mime-types that are possible to play in the web and check if the current gstreamer setup has support for them. Thus the backend won't traverse the whole gstreamer's registry per-call but rather a hash map to look up, just as in

https://github.com/WebKit/webkit/blob/0d960a0797d44e001af795b3d227be6b763fbbc0/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp#L2199

The return of the function perhaps should be an enum with Probably and Maybe

ferjm:
Thanks for the PR, @georgeroman 

I agree with @ceyusa, we should keep a list of mime-types that are supported on the web (we can get it from the WebKit's implementation that @ceyusa shared), lazily intersect that list with what's available in the GStreamer registry and cache the result of that intersection so subsequent calls to this function do not require traversing the entire registry again.

philn:
Please not the WebKit code linked here is obsolete, I implemented a new GstreamerRegistryScanner for media-capabilities support, it's used for canPlayType and the media-capabilities spec platform support.

ceyusa:
https://github.com/WebKit/webkit/blob/master/Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp

georgeroman:
I added some of the requested changes. I would really appreciate some feedback to see whether I'm heading in the right direction with this!

backends/gstreamer/lib.rs
```diff
@@ -49,12 +51,26 @@ use servo_media_streams::capture::MediaTrackConstraintSet;
...
 
-pub struct GStreamerBackend;
+pub struct GStreamerBackend {
+    registry_scanner: GStreamerRegistryScanner,
+}
+
+impl GStreamerBackend {
+    pub fn new() -> GStreamerBackend {
+        let mut registry_scanner = GStreamerRegistryScanner::new();
+        registry_scanner.initialize();
```
ferjm: Could we lazy initialize this?

backends/gstreamer/lib.rs
```diff
@@ -49,12 +51,26 @@ use servo_media_streams::capture::MediaTrackConstraintSet;
...
+
+        GStreamerBackend {
+            registry_scanner: registry_scanner,
```
ferjm: nit: we can use the field init shorthand here:

```rust
GStreamerBackend {
    registry_scanner,
}
```

backends/gstreamer/lib.rs
```diff
@@ -152,3 +187,246 @@ impl<'a> Drop for MutexGuard<'a> {
         }
     }
 }
+
+struct GStreamerRegistryScanner {
+    audio_decoder_factories: Vec<gst::ElementFactory>,
+    audio_parser_factories: Vec<gst::ElementFactory>,
+    video_decoder_factories: Vec<gst::ElementFactory>,
+    video_parser_factories: Vec<gst::ElementFactory>,
+    demux_factories: Vec<gst::ElementFactory>,
```
ferjm: Why do we need `*_factories` to be members of `GStreamerRegistryScanner`? It seems that we could have them as local variables for the `initialize` method.

backends/gstreamer/lib.rs
```diff
@@ -152,3 +188,246 @@ impl<'a> Drop for MutexGuard<'a> {
...
+            let is_vp8_decoder_available = has_element_for_media_type(&video_decoder_factories, "video/x-vp8");
+            let is_vp9_decoder_available = has_element_for_media_type(&video_decoder_factories, "video/x-vp9");
+
+            if is_vp8_decoder_available || is_vp9_decoder_available {
+                self.supported_mime_types.insert("video/webm");
+            }
+
+            if is_vp8_decoder_available {
+                self.supported_codecs.insert("vp8");
+                self.supported_codecs.insert("x-vp8");
```
philn: vp8.0 missing here

backends/gstreamer/lib.rs
```diff
@@ -152,3 +188,246 @@ impl<'a> Drop for MutexGuard<'a> {
...
+            }
+
+            if is_vp9_decoder_available {
+                self.supported_codecs.insert("vp9");
+                self.supported_codecs.insert("x-vp9");
```
philn: And vp9.0 here :)
I've fixed it in WebKit: https://bugs.webkit.org/show_bug.cgi?id=196615

backends/gstreamer/lib.rs
```diff
@@ -152,3 +188,248 @@ impl<'a> Drop for MutexGuard<'a> {
...
+            supported_mime_types: HashSet::new(),
+            supported_codecs: HashSet::new(),
+        }
+    }
+
+    pub fn is_container_type_supported(&mut self, container_type: &str) -> bool {
+        if !self.is_initialized {
+            self.initialize();
+            self.is_initialized = true;
+        }
```
ceyusa: what about using https://doc.rust-lang.org/std/sync/struct.Once.html for this, and a macro to call that once structure? :)

backends/gstreamer/lib.rs
```diff
@@ -49,12 +51,23 @@ use servo_media_streams::capture::MediaTrackConstraintSet;
...
 use servo_media_webrtc::{WebRtcBackend, WebRtcController, WebRtcSignaller};
+use std::collections::HashSet;
 
 lazy_static! {
     static ref BACKEND_BASE_TIME: gst::ClockTime = { gst::SystemClock::obtain().get_time() };
 }
 
-pub struct GStreamerBackend;
+pub struct GStreamerBackend {
+    registry_scanner: GStreamerRegistryScanner,
```
ferjm: Let's add some docs about what the `registry_scanner` is. i.e.: `The GStreamer registry holds the metadata of the set of plugins available in the host. This scanner is used to lazily analyze the registry and to provide information about the set of supported mime types and codecs that the backend is able to deal with`

backends/gstreamer/lib.rs
```diff
@@ -88,6 +101,29 @@ impl Backend for GStreamerBackend {
     fn create_videoinput_stream(&self, set: MediaTrackConstraintSet) -> Option<MediaStreamId> {
         media_capture::create_videoinput_stream(set)
     }
+
+    fn can_play_type(&mut self, media_type: &str) -> SupportsMediaType {
```
ceyusa: can_play_type() is a global backend function? or does it just belong to the player's API?

I am, actually, not sure.

ferjm: This will be useful for [MediaSource.isTypeSupported](https://developer.mozilla.org/en-US/docs/Web/API/MediaSource/isTypeSupported) as well, so it's ok to have it as a global backend function.

backends/gstreamer/lib.rs
ferjm: `lazy_static` takes care of this :)

ferjm:
Looks great! Thanks @georgeroman 

@bors-servo r+

PR #233, opened 2019-04-03
Streams registry

ferjm:


Manishearth:
Is there a reason the streams registry is only in the gstreamer backend? I'd imagine that the registry would be common to all backends.

There _will_ be client implementations of MediaStream eventually, for example. (since the DOM can write custom media streams)

Manishearth:
_eventually_ we'll probably want to switch this over to an event loop model, but that's only for when we move all of servo-media to its own process, which comes much later, and can be done pretty easily.

ferjm:
> Is there a reason the streams registry is only in the gstreamer backend? I'd imagine that the registry would be common to all backends.

No specific reason. It makes sense to share it among backends. I'll update the patch tomorrow. Thanks!

ferjm:
I've just updated the PR with the suggested change.

backends/dummy/lib.rs
```diff
@@ -36,24 +37,24 @@ impl BackendInit for DummyBackend {
 }
 
 impl Backend for DummyBackend {
-    fn create_audiostream(&self) -> Box<MediaStream> {
-        Box::new(DummyMediaStream)
+    fn create_audiostream(&self) -> MediaStreamId {
+        MediaStreamId::new()
```
Manishearth: Should the dummy backend be registering dummy streams instead? `register_stream(Box::new(DummyStream))`?

PR #236, opened 2019-04-06
 Render crates

ceyusa:
This PR creates two new crates inside gstreamer backend:

- A render crate which only defines a trait for different renders  for different OS 
- A render object helper that will setup the rendering bits for the player
- A unix-render crate which implementes the render trait for Unix-based GL systems

This will fix the problem with cargo with features per target.

It is not well tested yet. And the GL implementation still has some missing bits.

I would like to propose a new create depending on glutin to extract the GL context and display for different platforms used by the possible gl renders.

### DIFF
--- a/backends/gstreamer/player.rs
+++ b/backends/gstreamer/player.rs
@@ -426,7 +426,7 @@ impl GStreamerPlayer {
             .set_config(config)
             .map_err(|e| PlayerError::Backend(e.to_string()))?;
 
-        let render = GStreamerRender::new();
+        let render = GStreamerRender::new(GlContext::Unknown, 0);
         let appsink = render.setup_video_sink(&pipeline)?;
 
         // There's a known bug in gstreamer that may cause a wrong transition
--- a/backends/gstreamer/player.rs
+++ b/backends/gstreamer/player.rs
@@ -345,6 +345,7 @@ pub struct GStreamerPlayer {
     is_ready: Arc<Once>,
     /// Indicates whether the type of media stream to be played is a live stream.
     stream_type: StreamType,
+    render: RefCell<Option<GStreamerRender>>,
 }
 
 impl GStreamerPlayer {
@@ -355,6 +356,7 @@ impl GStreamerPlayer {
             renderers: Arc::new(Mutex::new(FrameRendererList::new())),
             is_ready: Arc::new(Once::new()),
             stream_type,
+            render: RefCell::new(None),
         }
     }
 
@@ -426,7 +428,10 @@ impl GStreamerPlayer {
             .set_config(config)
             .map_err(|e| PlayerError::Backend(e.to_string()))?;
 
-        let render = GStreamerRender::new(GlContext::Unknown, 0);
+        let render = self
+            .render
+            .replace(None)
+            .unwrap_or(GStreamerRender::new(GlContext::Unknown, 0));
         let appsink = render.setup_video_sink(&pipeline)?;
 
         // There's a known bug in gstreamer that may cause a wrong transition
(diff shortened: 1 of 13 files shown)


jdm:
Rather than adding a glutin dependency to the media crates, I think we should define a trait that provides the necessary APIs. This can then be implemented by the servo embedder, which will allow us to more easily integrate with platforms where we don't use glutin (like Android and magic leap).

ceyusa:
Just tested the branch in Mac and Win10, and in the first works, in the last compiles, but I got a strange webrender error. Removing WIP.

ceyusa:
@ferjm: r?

backends/gstreamer/player.rs
```diff
@@ -903,10 +788,15 @@ impl Player for GStreamerPlayer {
...
     }
 
-    fn set_gl_params(&self, _: GlContext, _: usize) -> Result<(), ()> {
-        // XXX All GL functionality is temporarily disabled because of
-        // https://github.com/servo/servo/pull/22944#issuecomment-468827665
-        Err(())
+    fn set_gl_params(&self, context: GlContext, display: usize) -> Result<(), ()> {
+        let render = GStreamerRender::new(context, display);
+        let ret = render.is_gl();
+        let _ = self.render.replace(Some(render));
```
ferjm:
```rust
*self.render.borrow_mut() = Some(render);
```

backends/gstreamer/player.rs
```diff
@@ -579,7 +461,7 @@ impl GStreamerPlayer {
         *self.inner.borrow_mut() = Some(Arc::new(Mutex::new(PlayerInner {
             player,
             source: None,
-            appsink: video_sink,
+            appsink: appsink,
```
ferjm: nit: we can use the field init shorthand now. `appsink` alone should work.

backends/gstreamer/player.rs
```diff
@@ -550,7 +428,11 @@ impl GStreamerPlayer {
             .set_config(config)
             .map_err(|e| PlayerError::Backend(e.to_string()))?;
 
-        let video_sink = self.setup_video_sink(&player)?;
+        let render = self
+            .render
+            .replace(None)
```
ferjm: I believe you want to use `into_inner` instead of `replace` here.

backends/gstreamer/render-unix/lib.rs
```diff
@@ -0,0 +1,193 @@
+#![cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
```
ferjm: TIL about DragonFly :)

backends/gstreamer/render-unix/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+extern crate gstreamer_gl as gst_gl;
+use gst_gl::prelude::*;
+extern crate gstreamer_video as gst_video;
+
+extern crate servo_media_gstreamer_render as sm_gst_render;
+use sm_gst_render::Render;
+
+extern crate servo_media_player as sm_player;
+use sm_player::frame::{Buffer, Frame, FrameData};
+use sm_player::{GlContext, PlayerError};
```
ferjm: uber-nit: could we group the `extern`s and the `use`s in separate blocks, please?

backends/gstreamer/render-unix/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+                        .get_property("context")
+                        .or_else(|_| Err(()))?
+                        .get::<gst_gl::GLContext>()
+                } else {
+                    None
+                };
+        }
+
+        //let sync_meta = buffer.as_ref().get_meta::<gst_gl::GLSyncMeta>().unwrap();
+        //sync_meta.set_sync_point(self.gst_context.borrow().as_ref().unwrap());
```
ferjm: I guess we can remove this commented code :)

backends/gstreamer/render-unix/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        appsink: &gst::Element,
+        pipeline: &gst::Element,
+    ) -> Result<(), PlayerError> {
+        if self.gl_upload.lock().unwrap().is_some() {
+            return Err(PlayerError::Backend(
+                "render unix already setup the video sink".to_owned(),
+            ));
+        }
+
+        let vsinkbin = gst::ElementFactory::make("glsinkbin", "servo-media-vsink")
```
ferjm: Would it make sense to add a check for this element in `set_gl_params` like the one that we have in [GStreamerPlayer.setup](https://github.com/servo/media/blob/master/backends/gstreamer/player.rs#L492)?

backends/gstreamer/player.rs
ceyusa: I get this error: 

```
error[E0507]: cannot move out of borrowed content
```

ceyusa:
@ferjm: r?

PR #239, opened 2019-04-18
Handle buffer offset in servosrc

ceyusa:
Seeking requires a strict handling of each buffer offset. Though appsrc do some offset handling, but only for live stream (not seeking). When random-access is enabled the application must take care of marking each buffer offset correctly.

This patch extends servosrc's method push_buffer() so it will count the current offset and stamps it on the pushed buffer. Also, it honors the buffer's block size defined in appsrc, splitting the input data in several buffers if it's bigger, fixing some problems with the buffering mechanism in playbin.

When a seek operation is done, the current media offset must be updated with the requested position.

As per media/player design, it is required that the user must push synchronized the data from the requested position by the seek.

### DIFF
--- a/examples/player/player_wrapper.rs
+++ b/examples/player/player_wrapper.rs
@@ -7,9 +7,10 @@ use servo_media::player::frame::{Frame, FrameRenderer};
 use servo_media::player::{GlContext, Player, PlayerEvent, StreamType};
 use servo_media::ServoMedia;
 use std::fs::File;
-use std::io::{BufReader, Read};
+use std::io::{BufReader, Read, Seek, SeekFrom};
 use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::mpsc;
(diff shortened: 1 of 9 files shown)


ceyusa:
@ferjm  r?

examples/player/player_wrapper.rs
```diff
@@ -70,53 +71,76 @@ impl PlayerWrapper {
...
+                                end_file.store(true, Ordering::Relaxed);
+                            }
+                            Ok(size) => {
+                                if let Err(e) = player
+                                    .lock()
+                                    .unwrap()
+                                    .push_data(Vec::from(&buffer[0..size]))
+                                {
+                                    println!("Can't push data: {:?}", e);
+                                    //break;
```
ferjm: Remove this comment?

ceyusa: Actually it is a hack left behind. Let bring back the break if it is not a enough data.

PR #244, opened 2019-04-30
Handle silent blocks in ChannelSplitterNode

Manishearth:


Manishearth:
This causes a panic in a gain node test (unsure how WPT hasn't caught it yet)

PR #245, opened 2019-05-03
Minor clean ups

ceyusa:
[ ] remove spurious variable mutability labels
[ ] have just one PlayerGLContext implementation in player test app


ceyusa:
@jdm r?

PR #246, opened 2019-05-06
Send panning message to stereo panner node.

jdm:
Otherwise this example panics.


Manishearth:
Hm, I thought I'd fixed this.

@bors-servo r+

Issue #249, opened 2019-05-07
Deadlock in webrtc stream acquisition


Manishearth:
Running simple_webrtc has a deadlock in stream acquisition during `link_stream()`

https://github.com/servo/media/blob/ccbf63f1a43e1525ae07b0cb21d1ee6e419ea139/backends/gstreamer/webrtc.rs#L381

cc @ferjm

ferjm:
Ugh, apologies for causing this regression. The webrtc example is not working for me, so I couldn't test it. Thanks for fixing it!

Manishearth:
No biggie. What about the example doesn't work? Is it just the signaling
server? I can add comments on how you're supposed to set that up

ferjm:
Last time I tried I was seeing this

```
0:00:22.921472000 11559 0x7fdb3f501490 ERROR glutils gstglutils.c:229:void gst_gl_element_propagate_display_context(GstElement *, GstGLDisplay *):<sink> Could not get GL display connection
```

And the example was crashing with a

```
[1] 11559 illegal hardware instruction GST_DEBUG=3 cargo ex simple_webrtc 8443.
```

I suspect there is something wrong with my gstreamer setup or something broken in webrtcbin and/or osxvideosink.

I am working on the support for multiple streams now, so I'll need to revisit this and try to make the example work for me.

Manishearth:
Two things: 

 - does the example work if you change simple_webrtc to only use videotestsrc (currently one side opens a capture stream)?
 - does the local pipeline of `videotestsrc ! autovideosink` work? It may need videoconvert/etc elements in between.

PR #250, opened 2019-05-07
Don't reentrantly lock the stream id

Manishearth:
fixes https://github.com/servo/media/issues/249

r? @jdm or @ferjm

### DIFF
--- a/backends/gstreamer/webrtc.rs
+++ b/backends/gstreamer/webrtc.rs
@@ -123,14 +123,16 @@ impl WebRtcControllerBackend for GStreamerWebRtcController {
     }
 
     fn add_stream(&mut self, stream_id: &MediaStreamId) -> WebrtcResult {
-        let stream =
-            get_stream(stream_id).expect("Media streams registry does not contain such ID");
-        let mut stream = stream.lock().unwrap();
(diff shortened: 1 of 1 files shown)


backends/gstreamer/webrtc.rs
```diff
@@ -123,14 +123,16 @@ impl WebRtcControllerBackend for GStreamerWebRtcController {
...
-        let stream =
-            get_stream(stream_id).expect("Media streams registry does not contain such ID");
-        let mut stream = stream.lock().unwrap();
-        let stream = stream
-            .as_mut_any()
-            .downcast_mut::<GStreamerMediaStream>()
-            .ok_or("Does not currently support non-gstreamer streams")?;
-        stream.insert_capsfilter();
+        {
+            let stream =
```
Manishearth: This does mean we end up locking twice. I could alternatively pass down the `&mut GstreamerMediaStream` to `link_stream` directly, but that gets uglier

Manishearth:
The second commit is optional, it avoids double-locking but the code is a bit more complicated

jdm:
@bors-servo r+
The complexity seems reasonable to me.

Issue #251, opened 2019-05-07
Build broken on Mac


ferjm:
```
error[E0277]: `*mut objc::runtime::Object` cannot be sent between threads safely
  --> examples/player/ui.rs:29:6
   |
29 | impl PlayerGLContext for PlayerContextGlutin {
   |      ^^^^^^^^^^^^^^^ `*mut objc::runtime::Object` cannot be sent between threads safely
   |
   = help: within `glutin::WindowedContext`, the trait `std::marker::Send` is not implemented for `*mut objc::runtime::Object`
   = note: required because it appears within the type `glutin::platform::platform::IdRef`
   = note: required because it appears within the type `glutin::platform::platform::WindowedContext`
   = note: required because it appears within the type `glutin::platform::platform::Context`
   = note: required because it appears within the type `glutin::Context`
   = note: required because it appears within the type `glutin::WindowedContext`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<glutin::WindowedContext>`
   = note: required because it appears within the type `ui::PlayerContextGlutin`

error[E0277]: `*mut objc::runtime::Object` cannot be shared between threads safely
  --> examples/player/ui.rs:29:6
   |
29 | impl PlayerGLContext for PlayerContextGlutin {
   |      ^^^^^^^^^^^^^^^ `*mut objc::runtime::Object` cannot be shared between threads safely
   |
   = help: within `glutin::WindowedContext`, the trait `std::marker::Sync` is not implemented for `*mut objc::runtime::Object`
   = note: required because it appears within the type `glutin::platform::platform::IdRef`
   = note: required because it appears within the type `glutin::platform::platform::WindowedContext`
   = note: required because it appears within the type `glutin::platform::platform::Context`
   = note: required because it appears within the type `glutin::Context`
   = note: required because it appears within the type `glutin::WindowedContext`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<glutin::WindowedContext>`
   = note: required because it appears within the type `ui::PlayerContextGlutin`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `examples`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

ceyusa:
https://github.com/ceyusa/media/commit/acff252b92317639c707acd4cae635f4c7dcd750

PR #253, opened 2019-05-07
Allow player example to build on macOS.

jdm:
Fixes #251.

PR #255, opened 2019-05-09
PlayerGLContext does not need to be thread aware

ceyusa:
Fixes: #251

As far as I have played, there's no need to make PlayerGLContext trait thread aware, so we could avoid the changes in https://github.com/servo/media/pull/253


ceyusa:
@ferjm r?

PR #256, opened 2019-05-09
Bump websocket version to 0.22

ceyusa:
Thus it will use a fixed version of openssl, avoiding this bug:

https://github.com/sfackler/rust-openssl/issues/987


ceyusa:
@Manishearth r?

PR #260, opened 2019-05-10
Allow simultaneous playback of audio and video streams

ferjm:


backends/gstreamer/media_stream_source.rs
```diff
@@ -50,21 +59,45 @@ mod imp {
...
+            let video_srcpad =
+                gst::GhostPad::new_no_target_from_template("video_stream_src", &pad_templ).unwrap();
+            flow_combiner.add_pad(&video_srcpad);
+
+            audio_srcpad.set_chain_function(|_, _, _| {
+                Ok(gst::FlowSuccess::Ok)
+            });
+
+            video_srcpad.set_chain_function(|pad, parent, buffer| {
+            //    let result = gst::ProxyPad::chain_default::<gst::GhostPad, ServoMediaStreamSrc, Option<ServoMediaStreamSrc>>(pad, Some(parent), buffer);
```
ferjm: @sdroege it's not clear to me how to use [chain_default](https://slomo.pages.freedesktop.org/rustdocs/gstreamer/gstreamer/struct.ProxyPad.html#method.chain_default) here, and I couldn't find any examples of it. Could you advice on which associated types should I use here? Thanks!

Manishearth:
How does setting multiple streams on this work? (or is that not implemented yet?) It seems like `set_stream` still gets called just once?

ferjm:
`set_stream` is called once per `MediaStreamTrack` in Servo https://github.com/servo/servo/blob/master/components/script/dom/htmlmediaelement.rs#L842

backends/gstreamer/media_stream_source.rs
sdroege: Why doesn't it compile with just
```rust
let result = gst::ProxyPad::chain_default(pad.downcast_ref::<gst::ProxyPad>().unwrap(), parent, buffer);
```

ferjm: Thanks! That worked

ferjm: Ugh, actually, it does not work.

```
error[E0283]: type annotations required: cannot resolve `_: glib::IsA<gst::Object>`
  --> backends/gstreamer/media_stream_source.rs:86:30
   |
86 |                 let result = gst::ProxyPad::chain_default(pad.upcast_ref::<gst::ProxyPad>(), parent, buffer);
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: required by `gst::proxy_pad::<impl gst::ProxyPad>::chain_default`
```

sdroege: `downcast_ref()`, not `upcast_ref()` :) You copied my code incorrectly. Not sure if *that* is the problem though. If what I pasted does not work please create an issue here https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/new

It's supposed to be used like that and if it can't then that's a problem

ferjm: Yeah, I tried `downcast_ref()` first, but it was throwing this build error:
```
error[E0277]: the trait bound `gst::ProxyPad: glib::IsA<gst::GhostPad>` is not satisfied
  --> backends/gstreamer/media_stream_source.rs:85:70
   |
85 |                 let chain_result = gst::ProxyPad::chain_default(&pad.downcast_ref::<gst::ProxyPad>(), parent, buffer);
   |                                                                      ^^^^^^^^^^^^ the trait `glib::IsA<gst::GhostPad>` is not implemented for `gst::ProxyPad`
   |
   = help: the following implementations were found:
             <gst::ProxyPad as glib::IsA<glib::Object>>
             <gst::ProxyPad as glib::IsA<gst::Object>>
             <gst::ProxyPad as glib::IsA<gst::Pad>>
             <gst::ProxyPad as glib::IsA<gst::ProxyPad>>
   = note: required because of the requirements on the impl of `glib::object::CanDowncast<gst::ProxyPad>` for `gst::GhostPad`
```

sdroege: Oh you have a ghostpad there already.

```rust
let chain_result = gst::ProxyPad::chain_default(pad, parent, buffer);
```

Does not work? It seems like everything already has the right type for you.

ferjm: No, it doesn't work.

```
error[E0283]: type annotations required: cannot resolve `_: glib::IsA<gst::Object>`
  --> backends/gstreamer/media_stream_source.rs:85:36
   |
85 |                 let chain_result = gst::ProxyPad::chain_default(pad, parent, buffer);
   |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: required by `gst::proxy_pad::<impl gst::ProxyPad>::chain_default`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0283`.
error: Could not compile `servo-media-gstreamer`.
```

sdroege: Ok, can you create an issue about this then I'll take a look :) Ideally with a small testcase, a patch to the `gst-plugin-tutorial/src/progressbin.rs` for example

sdroege: Maybe too generic code here :)

sdroege:
```rust
let chain_result = gst::ProxyPad::chain_default(pad, parent.as_ref(), buffer);
```

is the workaround for now btw. See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/203 and https://github.com/gtk-rs/gir/issues/777 for the effort to solve it in a nicer way everywhere else where this is wrong too.

backends/gstreamer/media_stream_source.rs
```diff
@@ -59,46 +122,20 @@ mod imp {
...
+        fn new_with_class(_: &subclass::simple::ClassStruct<Self>) -> Self {
             let audio_proxysrc = gst::ElementFactory::make("proxysrc", None)
                 .expect("Could not create proxysrc element");
-
-            let pad_templ = klass.get_pad_template("src").unwrap();
-            let audio_srcpad =
-                gst::GhostPad::new_no_target_from_template("audio_stream_src", &pad_templ).unwrap();
-            flow_combiner.lock().unwrap().add_pad(&audio_srcpad);
+            let audio_srcpad = gst::GhostPad::new_no_target_from_template(
+                "video_stream_src",
```
sdroege: Wrong name

backends/gstreamer/media_stream_source.rs
```diff
@@ -59,46 +122,20 @@ mod imp {
...
-            audio_internal_pad.set_chain_function(move |pad, parent, buffer| {
-                let chain_result = gst::ProxyPad::chain_default(pad, parent.as_ref(), buffer);
-                let result = combiner.lock().unwrap().update_pad_flow(pad, chain_result);
-                if result == Err(gst::FlowError::Flushing) {
-                    return chain_result;
-                }
-                result
-            });
+            let video_srcpad = gst::GhostPad::new_no_target_from_template(
+                "video_stream_src",
```
sdroege: Also wrong name :)

ferjm:
The meat of these patches is on the `backends/gstreamer/media_stream_source.rs` file. The rest is the addition of the `only_stream` flag to `set_stream` and rustfmt.

PR #271, opened 2019-06-21
global mute for servo-media backend

khodzha:
Issue description is here:
https://github.com/servo/servo/issues/22359

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -22,12 +22,15 @@ extern crate gstreamer_webrtc as gst_webrtc;
 extern crate ipc_channel;
 #[macro_use]
 extern crate lazy_static;
+#[macro_use]
+extern crate log;
 
 extern crate servo_media;
 extern crate servo_media_audio;
 extern crate servo_media_gstreamer_render;
 extern crate servo_media_player;
 extern crate servo_media_streams;
+extern crate servo_media_traits;
 extern crate servo_media_webrtc;
 extern crate url;
 
@@ -58,9 +61,12 @@ use servo_media_player::{Player, PlayerEvent, StreamType};
 use servo_media_streams::capture::MediaTrackConstraintSet;
 use servo_media_streams::registry::MediaStreamId;
 use servo_media_streams::MediaOutput;
+use servo_media_traits::{ClientContextId, Muteable};
 use servo_media_webrtc::{WebRtcBackend, WebRtcController, WebRtcSignaller};
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, Mutex};
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex, Weak};
+use std::vec::Vec;
 
 lazy_static! {
     static ref BACKEND_BASE_TIME: gst::ClockTime = { gst::SystemClock::obtain().get_time() };
@@ -68,26 +74,76 @@ lazy_static! {
 
 pub struct GStreamerBackend {
     capture_mocking: AtomicBool,
+    muteables: Mutex<HashMap<ClientContextId, Vec<(usize, Weak<Mutex<dyn Muteable>>)>>>,
+    next_muteable_id: AtomicUsize,
+}
+
+impl GStreamerBackend {
+    fn remove_muteable(&self, id: &ClientContextId, muteable_id: usize) {
+        let mut muteables = self.muteables.lock().unwrap();
+        if let Some(vec) = muteables.get_mut(&id) {
+            vec.retain(|m| m.0 != muteable_id);
+            if vec.len() == 0 {
+                muteables.remove(&id);
+            }
+        }
+    }
 }
 
 impl Backend for GStreamerBackend {
     fn create_player(
         &self,
+        id: &ClientContextId,
         stream_type: StreamType,
         sender: IpcSender<PlayerEvent>,
         renderer: Option<Arc<Mutex<dyn FrameRenderer>>>,
         gl_context: Box<dyn PlayerGLContext>,
-    ) -> Box<dyn Player> {
-        Box::new(player::GStreamerPlayer::new(
+    ) -> Arc<Mutex<dyn Player>> {
+        let muteable_id = self.next_muteable_id.fetch_add(1, Ordering::Relaxed);
+        let player = Arc::new(Mutex::new(player::GStreamerPlayer::new(
+            muteable_id,
             stream_type,
             sender,
             renderer,
             gl_context,
-        ))
+        )));
+        let mut muteables = self.muteables.lock().unwrap();
+        let entry = muteables.entry(*id).or_insert(Vec::new());
+        entry.push((muteable_id, Arc::downgrade(&player).clone()));
+        player
     }
 
-    fn create_audio_context(&self, options: AudioContextOptions) -> AudioContext {
-        AudioContext::new::<Self>(options)
+    fn shutdown_player(&self, id: &ClientContextId, player: Arc<Mutex<dyn Player>>) {
+        let player = player.lock().unwrap();
+        let p_id = player.get_id();
+        self.remove_muteable(id, p_id);
+
+        if let Err(e) = player.shutdown() {
+            warn!("Player was shut down with err: {:?}", e);
+        }
+    }
+
+    fn create_audio_context(
+        &self,
+        id: &ClientContextId,
+        options: AudioContextOptions,
+    ) -> Arc<Mutex<AudioContext>> {
+        let muteable_id = self.next_muteable_id.fetch_add(1, Ordering::Relaxed);
+        let context = Arc::new(Mutex::new(AudioContext::new::<Self>(muteable_id, options)));
+        let mut muteables = self.muteables.lock().unwrap();
+        let entry = muteables.entry(*id).or_insert(Vec::new());
+        entry.push((muteable_id, Arc::downgrade(&context).clone()));
+        context
+    }
+
+    fn shutdown_audio_context(
+        &self,
+        id: &ClientContextId,
+        audio_context: Arc<Mutex<AudioContext>>,
+    ) {
+        let audio_context = audio_context.lock().unwrap();
+        let ac_id = audio_context.get_id();
+        self.remove_muteable(id, ac_id);
     }
 
     fn create_webrtc(&self, signaller: Box<dyn WebRtcSignaller>) -> WebRtcController {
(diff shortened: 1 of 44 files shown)


ceyusa:
Thanks @khodzha! 

It is a big patch.

Would you mind to explain or describe a bit your proposed solution? :)

player/Cargo.toml
```diff
@@ -16,3 +16,6 @@ ipc-channel = "0.11"
 
 [dependencies.servo-media-streams]
 path = "../streams"
+
+[dependencies.servo-media-audio]
+path = "../audio"
```
ferjm: Using a `servo_media_traits` crate, we won't need this anymore.

servo-media/lib.rs
```diff
@@ -28,20 +29,22 @@ pub trait BackendInit {
...
     fn create_videostream(&self) -> MediaStreamId;
     fn create_stream_output(&self) -> Box<dyn MediaOutput>;
     fn create_audioinput_stream(&self, set: MediaTrackConstraintSet) -> Option<MediaStreamId>;
     fn create_videoinput_stream(&self, set: MediaTrackConstraintSet) -> Option<MediaStreamId>;
-    fn create_audio_context(&self, options: AudioContextOptions) -> AudioContext;
+    fn create_audio_context(&self, id: MediaContextId, options: AudioContextOptions) -> Arc<Mutex<AudioContext>>;
     fn create_webrtc(&self, signaller: Box<dyn WebRtcSignaller>) -> WebRtcController;
     fn can_play_type(&self, media_type: &str) -> SupportsMediaType;
     fn set_capture_mocking(&self, _mock: bool) {}
+    fn mute(&self, _val: bool) {}
```
ferjm: nit: this should probably be named `set_mute` to make it consistent with the `Mutable::set_mute` function.

Some documentation would also be really appreciated (i.e. mention that this method mute/unmute all Player and AudioContext instances).

examples/android/lib/src/lib.rs
```diff
@@ -13,42 +13,51 @@ use servo_media::audio::node::AudioScheduledSourceNodeMessage;
...
 use servo_media::{ServoMedia};
+#[cfg(target_os = "android")]
+use servo_media::audio::muteable::MediaContextId;
+#[cfg(target_os = "android")]
+use std::sync::{Arc, Mutex};
 
 #[cfg(target_os = "android")]
 struct AudioStream {
-    context: AudioContext,
+    context_mutex: Arc<Mutex<AudioContext>>,
```
ferjm: nit: leave the previous naming, please: `context`.

examples/audio_decoder.rs
```diff
@@ -12,9 +12,11 @@ use std::path::Path;
 use std::sync::mpsc;
 use std::sync::{Arc, Mutex};
 use std::{thread, time};
+use servo_media::audio::muteable::MediaContextId;
 
 fn run_example(servo_media: Arc<ServoMedia>) {
-    let context = servo_media.create_audio_context(Default::default());
+    let context_mutex = servo_media.create_audio_context(MediaContextId::build(1, 1), Default::default());
```
ferjm: Same thing about the naming: `context`.

examples/biquad.rs
```diff
@@ -10,9 +10,11 @@ use servo_media::audio::param::{ParamType, RampKind, UserAutomationEvent};
 use servo_media::ServoMedia;
 use std::sync::Arc;
 use std::{thread, time};
+use servo_media::audio::muteable::MediaContextId;
 
 fn run_example(servo_media: Arc<ServoMedia>) {
-    let context = servo_media.create_audio_context(Default::default());
+    let context_mutex = servo_media.create_audio_context(MediaContextId::build(1, 1), Default::default());
```
ferjm: Ditto

examples/channels.rs
```diff
@@ -8,9 +8,11 @@ use servo_media::audio::oscillator_node::OscillatorNodeOptions;
 use servo_media::ServoMedia;
 use std::sync::Arc;
 use std::{thread, time};
+use servo_media::audio::muteable::MediaContextId;
 
 fn run_example(servo_media: Arc<ServoMedia>) {
-    let context = servo_media.create_audio_context(Default::default());
+    let context_mutex = servo_media.create_audio_context(MediaContextId::build(1, 1), Default::default());
```
ferjm: Same thing :) which applies to the rest of examples.

servo-media/lib.rs
khodzha: is leaving `mute` fn name in `Muteable` trait ok?
`set_mute` will conflict with `inner_player_proxy!(set_mute, val, bool);` of `impl Player for GStreamerPlayer`

khodzha:
@ceyusa its basically an implementation of changes proposed [here](https://github.com/servo/servo/issues/22359#issuecomment-445038353) with some caveats (related to servo-media updates) figured out

servo-media/lib.rs
ferjm: Yes, it's not a big deal. And actually, I was wrong asking this. I don't know why I thought `Muteable` had a `set_mute` function, but it is actually named `mute`. So just leave both methods as `mute`.

In any case, if you find yourself again in this scenario where you have to call methods with the same name, you can use https://doc.rust-lang.org/stable/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name

examples/audio_decoder.rs
khodzha: could you elaborate on this case?
there is an actual `context`, which is locked & unwrapped from `context_mutex` at the next line; also i need to bind `context_mutex` since it needs to live till the end of `run_example`

would renaming `context_mutex` -> `context` and `context` -> `locked_context` be ok?

ferjm: There is no need to use a different variable name in this case. Rust provides a way to [shadow variables](https://doc.rust-lang.org/book/ch03-01-variables-and-mutability.html#shadowing).

```rust
    let context = servo_media.create_audio_context(MediaContextId::build(1, 1), Default::default());
    let context = context.lock().unwrap();
```

khodzha:
Shouldnt `mute()` receive ClientContextId argument? :thinking: 

Because if my understanding is correct and all Player/AudioContext instances on a single tab have same ClientContextId (BrowsingContextID) of this tab and we want to mute a single tab and not all tabs, we need to mute only muteables with given ClientContextId

ferjm:
Of course, it should! Sorry, I missed that on my review.

backends/gstreamer/lib.rs
```diff
@@ -58,36 +59,50 @@ use servo_media_player::{Player, PlayerEvent, StreamType};
...
 use std::sync::{Arc, Mutex};
+use std::vec::Vec;
 
 lazy_static! {
     static ref BACKEND_BASE_TIME: gst::ClockTime = { gst::SystemClock::obtain().get_time() };
 }
 
 pub struct GStreamerBackend {
     capture_mocking: AtomicBool,
+    muteables: Mutex<HashMap<ClientContextId, Vec<Arc<Mutex<dyn Muteable>>>>>,
```
ferjm: We need a way to remove `Muteable`s from this map when they are dropped.

Here's an idea to fix this:

- Add `shutdown_player(&self, client_context_id: &ClientContextId, player: Arc<Mutex<dyn Player>)` and `shutdown_audio_context(&self, client_context_id: &ClientContextId, audio_context: Arc<Mutex<AudioContext>>)` methods to the `Backend` trait.
- For the GStreamer implementation of these methods, remove the `Player` or `AudioContext` being shutdown from the `muteables` map. Note that if the `Vec` is empty after removing the `muteable`, we should remove the client context entry as well. We should also call the `Player.shutdown()` method within the `shutdown_player` implementation.
- To make sure that we don't keep expired instances alive if the client forgets to call these `shutdown_*` methods, one thing that we can do here is to use `Weak` instead of `Arc`. So: `muteables: Mutex<HashMap<ClientContextId, Vec<Weak<Mutex<dyn Muteable>>>>>,`.

traits/lib.rs
```diff
@@ -0,0 +1,15 @@
+use std::num::NonZeroU32;
```
ferjm: Add the MPL2.0 license header here, please.

traits/muteable.rs
```diff
@@ -0,0 +1,3 @@
+pub trait Muteable: Send {
```
ferjm: I think we can add this to `traits/lib.rs` directly.

backends/gstreamer/lib.rs
```diff
@@ -22,12 +22,14 @@ extern crate gstreamer_webrtc as gst_webrtc;
 extern crate ipc_channel;
 #[macro_use]
 extern crate lazy_static;
+extern crate log;
```
ferjm:
```rust
#[macro_use]
extern crate log;
```

backends/gstreamer/lib.rs
```diff
@@ -44,6 +46,7 @@ pub mod webrtc;
 
 use gst::ClockExt;
 use ipc_channel::ipc::IpcSender;
+use log::warn;
```
ferjm: Importing the crate with `#[macro_use]` we don't need to add this `use`.

backends/gstreamer/lib.rs
```diff
@@ -58,36 +61,90 @@ use servo_media_player::{Player, PlayerEvent, StreamType};
 use servo_media_streams::capture::MediaTrackConstraintSet;
 use servo_media_streams::registry::MediaStreamId;
 use servo_media_streams::MediaOutput;
+use servo_media_traits::Muteable;
+use servo_media_traits::ClientContextId;
```
ferjm: nit: alphabetically ordered, please. Running `rustfmt` should fix this.

backends/gstreamer/lib.rs
```diff
@@ -58,36 +61,90 @@ use servo_media_player::{Player, PlayerEvent, StreamType};
...
-        Box::new(player::GStreamerPlayer::new(
-            stream_type,
-            sender,
-            renderer,
-            gl_context,
-        ))
+    ) -> Arc<Mutex<dyn Player>> {
+        let muteable_id = self.next_muteable_id.fetch_add(1, Ordering::Relaxed);
+        let player = player::GStreamerPlayer::new(muteable_id, stream_type, sender, renderer, gl_context);
+        let wrapped_player = Arc::new(Mutex::new(player));
```
ferjm: nit: let's not use an intermediate variable here.
```rust
let player = Arc::new(Mutex::new(player::GStreamerPlayer::new(muteable_id, stream_type, sender, renderer, gl_context)));
```

backends/gstreamer/lib.rs
```diff
@@ -58,36 +61,90 @@ use servo_media_player::{Player, PlayerEvent, StreamType};
...
+    ) {
+        let mut muteables = self.muteables.lock().unwrap();
+        if let Some(vec) = muteables.get_mut(&id) {
+            let audio_context = audio_context.lock().unwrap();
+            let ac_id = audio_context.get_id();
+            vec.retain(|m| m.0 != ac_id);
+            if vec.len() == 0 {
+                muteables.remove(&id);
+            }
+        }
```
ferjm: Could we move this code to a function (i.e. remove_mutable) and use it from `shutdown_player` and `shutdown_audio_context`?

backends/gstreamer/lib.rs
```diff
@@ -159,6 +216,23 @@ impl Backend for GStreamerBackend {
     fn set_capture_mocking(&self, mock: bool) {
         self.capture_mocking.store(mock, Ordering::Release)
     }
+
+    fn mute(&self, id: &ClientContextId, val: bool) {
+        let muteables = self.muteables.lock().unwrap();
+        match muteables.get(id) {
+            Some(vec) => vec.into_iter().for_each(|(_muteable_id, weak)| {
+                if let Some(mutex) = weak.upgrade() {
```
ferjm: If `upgrade` returns `None` it means that the `Mutable` has since been dropped. In that case we should remove it from the vec.

khodzha:
not sure if implementing `remove_mutable` straight on GStreamerBackend is ok :thinking:

ferjm:
> not sure if implementing `remove_mutable` straight on GStreamerBackend is ok 🤔
> (i guess it is but ¯_(ツ)_/¯)

Why wouldn't it be ok? :)

khodzha:
thank you for bearing with me <3

PR #272, opened 2019-06-26
Update to gstreamer-rs 0.14

ceyusa:


backends/gstreamer/audio_decoder.rs
```diff
@@ -220,7 +218,7 @@ impl AudioDecoder for GStreamerAudioDecoder {
                                     })?;
 
                                     for position in positions.iter() {
-                                        let buffer = buffer.clone();
+                                        let buffer = buffer.to_owned();
```
sdroege: Why is this necessary, a whole buffer copy instead of just getting a new reference?

backends/gstreamer/render.rs
```diff
@@ -96,9 +96,9 @@ impl GStreamerRender {
     }
 
     pub fn get_frame_from_sample(&self, sample: &gst::Sample) -> Result<Frame, ()> {
-        let buffer = sample.get_buffer().ok_or_else(|| ())?;
+        let buffer = sample.get_buffer().ok_or_else(|| ())?.to_owned();
```
sdroege: Same question here

sdroege:
Generally it looks like the update improved the code a bit and made it nicer to read. Or what do you think? :)

backends/gstreamer/audio_decoder.rs
sdroege: There's `Sample::get_buffer_owned()` if that's the issue here, which allows to do this without a copy of the buffer. See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/204 for the background.

ceyusa:
> Generally it looks like the update improved the code a bit and made it nicer to read. Or what do you think? :)

Agree, it feels more rusty, in the sense of the language, of course :)

Thanks @sdroege, specially for the `get_buffer_owned()` hint!

Manishearth:
@bors-servo r+

Should also fix the unsafe code used in webrtcbin for https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/189

PR #275, opened 2019-07-05
Shutdown audio contexts and players in all the examples

ferjm:


ceyusa:
lgtm

@bors-servo r+

jdm:
Out of curiosity, what is the current behaviour without these changes?

ceyusa:
> Out of curiosity, what is the current behaviour without these changes?

AFAIK the media elements in the examples are not removed from the global hashmaps.

Manishearth:
Why not have a destructor on the context handle that sends a shutdown message on death?

ferjm:
> > Out of curiosity, what is the current behaviour without these changes?
> 
> AFAIK the media elements in the examples are not removed from the global hashmaps.

Essentially that. Although these are weak references to the media elements. And they may be removed on following mutes/unmutes.

> Why not have a destructor on the context handle that sends a shutdown message on death?

Could you elaborate on this, please? We can improve this by making the Player/AudioContext send such a message to the owner GStreamerBackend on drop. I am not sure if that's what you meant.

ceyusa:
Just to clarify, the Backend, once instantiated, it is kept alive until the browser shutdowns, then we should remove the media elements when they are removed of the webpage while the browser keeps running.

ceyusa:
bors-servo ignores me :(

Well, perhaps it is better to clarify these questions before merging :)

ferjm:
Closing in favor of #279

Issue #259, opened 2019-05-10
Enhance the arguments handling in player example


ceyusa:
As @philn [pointed out](https://github.com/servo/media/pull/257#discussion_r282561592) we shall enhance the handling of arguments in the player test.

Perhaps using [clap](https://clap.rs/)

brightly-salty:
I think this should be closed? It looks like it was fixed by PR#277. If not, and it's still an issue, I would be willing to attempt it.

jdm:
Looks like it. Thanks!

PR #277, opened 2019-07-06
delegate player example args parsing to clap

khodzha:
Issue #259.

I'm not sure if `--no-video` and `--gl` help strings are valid though.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 license = "MPL-2.0"
 
 [dependencies]
+clap = { version = "2.33", default-features = false }
 euclid = "0.19.0"
 gleam = "0.6.8"
 hyper = "0.12"
--- a/examples/player/main.rs
+++ b/examples/player/main.rs
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+#[cfg(not(target_os = "android"))]
+extern crate clap;
 extern crate gleam;
 #[cfg(not(target_os = "android"))]
 extern crate glutin;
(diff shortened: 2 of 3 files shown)


ceyusa:
I already have a patch for this but "forgot" it because I'm rewriting the player example.

I should change that a push these things before duplicating work.

examples/Cargo.toml
```diff
@@ -20,6 +20,10 @@ webrender = "0.58.0"
 websocket = "0.22"
 ipc-channel = "0.11"
 
+[dependencies.clap]
+version = "2.33"
+default-features = false
```
ceyusa: why are you disabling the default features?

I would just add ```clap = "2.33"``` in the dependencies section in alphabetic order.

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
-            (true, false, fname)
-        } else {
-            panic!("Usage: cargo run --bin player [[--gl]|[--no-video]] <file_path>")
-        }
-    } else {
-        panic!("Usage: cargo run --bin player [--gl] <file_path>")
-    };
+    let clap_matches = clap::App::new("Servo-media player example")
+        .setting(clap::AppSettings::DisableVersion)
+        .usage("player [--gl|--no-video] <FILE>")
```
ceyusa: The autogenerated usage string should be good enough, isn't it?

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
+        .arg(
+            clap::Arg::with_name("gl")
+                .long("gl")
+                .display_order(1)
```
ceyusa: is display_order necessary?

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
-            let fname: &str = args[2].as_ref();
-            (true, false, fname)
-        } else {
-            panic!("Usage: cargo run --bin player [[--gl]|[--no-video]] <file_path>")
-        }
-    } else {
-        panic!("Usage: cargo run --bin player [--gl] <file_path>")
-    };
+    let clap_matches = clap::App::new("Servo-media player example")
+        .setting(clap::AppSettings::DisableVersion)
```
ceyusa: nice!

Though I added 

```
.author("Servo developers")
.about("Servo/MediaPlayer example using WebRender")
```

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
+        .usage("player [--gl|--no-video] <FILE>")
+        .arg(
+            clap::Arg::with_name("gl")
+                .long("gl")
+                .display_order(1)
+                .help("use opengl"),
```
ceyusa: In my version:

```
.help("Tries to render frames as GL textures")
.conflicts_with("NOVIDEO"),
```

The conflict_with is important.

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
+        )
+        .arg(
+            clap::Arg::with_name("no-video")
+                .long("no-video")
+                .display_order(2)
+                .help("dont render video"),
```
ceyusa:
```
 .help("Don't render video, only audio"),
```

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
+        )
+        .arg(
+            clap::Arg::with_name("file")
+                .required(true)
+                .value_name("FILE")
+                .display_order(3),
+        )
+        .group(clap::ArgGroup::with_name("use_gl").args(&["gl", "no-video"]))
```
ceyusa: This grouping is required?

examples/player/main.rs
```diff
@@ -283,23 +285,33 @@ fn main() {
...
+        .get_matches();
+
+    let no_video = clap_matches.is_present("no-video");
+    let use_gl = clap_matches.is_present("gl");
+    let filename = clap_matches.value_of("file").unwrap();
```
ceyusa:
```
+    let path = matches
+        .value_of("FILE")
+        .and_then(|s| Some(Path::new(s)))
+        .unwrap();
 
-    let path = Path::new(filename);
```

examples/Cargo.toml
khodzha: to be honest i was assuming you wouldnt want dependecies bloat and since these features arent critical i can just disable them

default features are suggestions (for typos), color for coloring errors and vecmap for a slight performance, you can see them [here](https://github.com/clap-rs/clap#optional-dependencies--features)

examples/player/main.rs
khodzha: Maybe its my fault but the default usage string I got from clap here was:
`<FILE> <--gl|--no-video>` implying either of `gl` or `no-video` is required.
So this was the solution i went for.

Maybe it could be handled with `args` tuning but I couldn't find a way.

examples/player/main.rs
khodzha: No, it just reorders them in help message, bringing default `--help` flag to the bottom, looked fancier to me.

examples/player/main.rs
khodzha: It's an alternative to `.conflicts_with("NOVIDEO")`, so I suppose its just a matter of preference

khodzha:
also sorry for kicking this under you, i should've commented in the issue first

examples/Cargo.toml
ceyusa: I saw the Cargo.lock diff and I got your point.

Though, as Captain Nitpick, I'd still change this to

```
[dependencies]
clap = { version = "2.33", default-features = false }
env_logger = "0.5"
```

examples/player/main.rs
ceyusa: Agreed

examples/player/main.rs
ceyusa: Agreed

ceyusa:
lgtm!

PR #278, opened 2019-07-07
WIP: replaced Vec with HashSet in gstreamer muteables

khodzha:
@ferjm you've been suggesting HashSet from the start yet because we didn't have an id back then I went with Vec.

But since we already add usize ids to muteables Vec I suppose we can just utilize it for Hash + Eq and replace Vec with HashSet now.

(I'm not sure whether IdMuteable is appropriate naming at all and also whether it should go straight to `lib.rs` file or be separated into another one)

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -63,28 +63,50 @@ use servo_media_streams::registry::MediaStreamId;
 use servo_media_streams::MediaOutput;
 use servo_media_traits::{ClientContextId, Muteable};
 use servo_media_webrtc::{WebRtcBackend, WebRtcController, WebRtcSignaller};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::hash::{Hash, Hasher};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Weak};
-use std::vec::Vec;
 
 lazy_static! {
     static ref BACKEND_BASE_TIME: gst::ClockTime = { gst::SystemClock::obtain().get_time() };
 }
 
+struct IdMuteable {
+    id: usize,
+    muteable: Weak<Mutex<dyn Muteable>>,
+}
+impl IdMuteable {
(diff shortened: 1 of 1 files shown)


backends/gstreamer/lib.rs
```diff
@@ -63,28 +63,50 @@ use servo_media_streams::registry::MediaStreamId;
...
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Weak};
-use std::vec::Vec;
 
 lazy_static! {
     static ref BACKEND_BASE_TIME: gst::ClockTime = { gst::SystemClock::obtain().get_time() };
 }
 
+struct IdMuteable {
+    id: usize,
```
ferjm: `Player`s and `AudioContext`s already have their own ids. I think there is no need to duplicated them here.

backends/gstreamer/lib.rs
```diff
@@ -63,28 +63,50 @@ use servo_media_streams::registry::MediaStreamId;
...
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Weak};
-use std::vec::Vec;
 
 lazy_static! {
     static ref BACKEND_BASE_TIME: gst::ClockTime = { gst::SystemClock::obtain().get_time() };
 }
 
+struct IdMuteable {
```
ferjm: nit: Let's rename this to `MuteableRef`.

backends/gstreamer/lib.rs
khodzha: yes, but we have only `Weak<Mutex<dyn Muteable>>`, is locking mutex just to get an id fine?

ferjm: Hmm, I think that should be fine.

khodzha: ah, this id is also used for Eq + Hash!

khodzha:
After a discussion with @ferjm it was decided HashSet doesnt provide much of a benefit since in current implementation `retain()` loop cant be replaced with a `remove()` (which was the idea behind using HashSet in the first place).

So I'm closing this PR.

PR #279, opened 2019-07-08
Remove shutdown_player and shutdown_audio_context

ferjm:
This is a cleaner approach that does not require consumers of the API to call any shutdown* method to ensure that the Muteables are removed appropriately.

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -61,12 +61,14 @@ use servo_media_player::{Player, PlayerEvent, StreamType};
 use servo_media_streams::capture::MediaTrackConstraintSet;
 use servo_media_streams::registry::MediaStreamId;
 use servo_media_streams::MediaOutput;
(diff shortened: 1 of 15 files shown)


backends/gstreamer/lib.rs
```diff
@@ -106,80 +110,80 @@ impl GStreamerBackend {
...
+                    BackendMsg::Shutdown(context_id, muteable_id) => {
+                        if let Some(vec) = instances_.lock().unwrap().get_mut(&context_id) {
+                            vec.retain(|m| m.0 != muteable_id);
+                            if vec.is_empty() {
+                                instances_.lock().unwrap().remove(&context_id);
+                            }
+                        }
+                    }
+                };
+            })
```
ceyusa: how do we stop this thread?

ferjm: The thread dies as soon as it processes the `BackendMsg::Shutdown` channel message.

PR #281, opened 2019-07-10
code clean-ups

ceyusa:
In order to avoid stepping on each other's toes, I send this pull request with a couple patches that clean up a bit our code, which I keep in under my working branches :)


ceyusa:
@ferjm r?

PR #283, opened 2019-07-10
Disable progressive downloading on Windows.

jdm:
Background is in https://github.com/servo/servo/issues/23705#issuecomment-510127043 and https://github.com/servo/media/issues/282. With this change I'm able to play various pieces of media in both UWP and desktop Servo builds.


jdm:
Given the small number of people working on Windows-specific things at the moment, my preference is to disable the functionality to allow the UWP work to make progress, rather than waiting to figure out an optimal fix.

PR #284, opened 2019-07-17
[WIP] GL support for MacOS

ferjm:


backends/gstreamer/render/Cargo.toml
```diff
@@ -10,6 +10,9 @@ path = "lib.rs"
 [dependencies.gstreamer]
 version = "0.14"
 
+[dependencies.gstreamer-gl]
+version = "0.14"
+
```
ceyusa: This is problematic. I had to split in crates the renders because, for some platforms it is not possible to link with gstreamer-gl. Thus this basic render crate shouldn't depende on gstreamer-gl.

ferjm:
Closing in favour of #315

PR #289, opened 2019-07-25
Fix race condition with seek lock by enforcing an ack 

ferjm:


backends/gstreamer/player.rs
```diff
@@ -664,7 +669,8 @@ impl GStreamerPlayer {
                                                 seek_channel.lock().unwrap().sender()
                                             )
                                         );
-                                        let ret = seek_channel.lock().unwrap().await();
+                                        let (ret, ack_channel) = seek_channel.lock().unwrap().await();
+                                        ack_channel.send(()).unwrap();
```
ceyusa: shouldn't this `ack_channel.send(()).unwrap()` be after `servosrc_.set_seek_done()`???

ceyusa: I was thinking, rather than juggling with another ipc channel, what about sending a SeekDone message after the `set_seek_done()`

PR #290, opened 2019-07-29
Add feature handling

ceyusa:
These two patches add

* Split the different dependencies for samples, keeping a control of which dependencies are required by which examples.
* Added a GL winsys feature, for unix render, so it can modified for different CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ matrix:
 
       script:
         - cargo build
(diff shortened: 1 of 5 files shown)


ceyusa:
@ferjm what do you think about this approach?

jdm:
Redirecting to @manishearth since ferjm is on vacation.

Manishearth:
I like the idea. I don't think it's super necessary and it may be additional friction when working on it, but overall should be a plus.

Make sure you update Servo to enable the various features.

@bors-servo r+

PR #293, opened 2019-08-19
Implement missing functionality of AudioBufferSourceNode

collares:
This is most of the fix for [Servo's #22363](https://github.com/servo/servo/issues/22363). It is broken into four pieces. Let me know if I can help in the reviewing process.

### DIFF
--- a/audio/buffer_source_node.rs
+++ b/audio/buffer_source_node.rs
@@ -306,8 +306,7 @@ impl AudioNodeEngine for AudioBufferSourceNode {
             inputs.blocks.push(block);
         }
 
-        if self.buffer_pos < 0.
-            || self.buffer_pos >= buffer.len() as f64
+        if !self.loop_enabled && (self.buffer_pos < 0. || self.buffer_pos >= buffer.len() as f64)
             || self.buffer_duration <= 0.
         {
             self.maybe_trigger_onended_callback();
--- a/audio/buffer_source_node.rs
+++ b/audio/buffer_source_node.rs
@@ -8,8 +8,8 @@ use param::{Param, ParamType};
 pub enum AudioBufferSourceNodeMessage {
     /// Set the data block holding the audio sample data to be played.
     SetBuffer(Option<AudioBuffer>),
-    /// Set start parameters (offset, duration).
-    Start(Option<f64>, Option<f64>),
+    /// Set start parameters (when, offset, duration).
+    Start(f64, Option<f64>, Option<f64>),
 }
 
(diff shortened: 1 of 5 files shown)


audio/buffer_source_node.rs
```diff
@@ -61,12 +70,19 @@ pub(crate) struct AudioBufferSourceNode {
...
     playback_rate: Param,
     /// Time at which the source should start playing.
     start_at: Option<Tick>,
+    /// Offset parameter passed to Start().
+    start_offset: Option<f64>,
+    /// Duration parameter passed to Start().
+    start_duration: Option<f64>,
+    /// The same as start_at, but with subsample accuracy.
+    /// FIXME: Maybe AudioScheduledSourceNode should use this as well?
+    start_when: f64,
```
ferjm: Yes, please. Let's make this part of `AudioScheduledSourceNode`. It could be done in a follow-up though.

audio/buffer_source_node.rs
```diff
@@ -8,6 +8,8 @@ use param::{Param, ParamType};
 pub enum AudioBufferSourceNodeMessage {
     /// Set the data block holding the audio sample data to be played.
     SetBuffer(Option<AudioBuffer>),
+    /// Set start parameters (when, offset, duration).
+    Start(f64, Option<f64>, Option<f64>),
```
ferjm: nit: rename this to `SetStartParams` or such, so it is not confused with the `AudioScheduledSourceNodeMessage::Start` message.

It would also be nice to remove the first parameter (`when`) as we can get it from `AudioScheduledSourceNodeMessage::Start`.

collares: I fixed the first part, and will leave the second part to the AudioScheduledSourceNode follow-up.

collares:
Addressed review comments! There are two remaining relevant WPT failures:

1) I have no idea what's happening with sub-sample-buffer-stitching.html. If I force the test to run, we seem to be huge values at the end of the test (of the order of 1e20), but the test only contanis samples in the range [-1, 1]. I don't see how any part of my code could generate such large values.

2) buffer-resampling.html fails because we interpolate the last sample with 0 when we output a sample corresponding to buffer.len() - 1 < buffer_pos < buffer.len(). The test wants us to hold the last sample constant for such values of buffer_pos, but it is unclear to me if this is the right thing to do.

If you are OK with the above failures, this PR is ready to be committed.

ferjm:
Thanks! Could you file follow-up issues for the remaining work (i.e. all the XXX comments that you added to the code) so we don't lose track of it, please?

@bors-servo r+

PR #302, opened 2019-08-22
AudioBufferSourceNode: Hold last sample instead of interpolating with 0

### LINKED ISSUE #300: AudioBufferSourceNode: Decide if we should interpolate with silence after last sample
buffer-resampling.html (WPT test) fails because we interpolate the last sample with 0 when we output a sample with pos corresponding to buffer.len() - 1 < pos < buffer.len(). The test wants us to hold the last sample constant for such values of pos, but it is unclear to me if this is the right thing to do. 

The spec should be clarified, in my opinion. I will file a spec bug. The relevant code is https://github.com/servo/media/blob/86b9a2892af3774e42243ac84394a052719b2bf6/audio/buffer_source_node.rs#L399

collares:
Fixes #300 now that the spec has been clarified.

### DIFF
--- a/audio/buffer_source_node.rs
+++ b/audio/buffer_source_node.rs
@@ -127,9 +127,7 @@ impl AudioBufferSourceNode {
             AudioBufferSourceNodeMessage::SetLoopEnabled(loop_enabled) => {
                 self.loop_enabled = loop_enabled
             }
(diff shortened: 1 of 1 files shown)


collares:
Closing this for the time being. See https://github.com/WebAudio/web-audio-api/issues/2032#issuecomment-524120143 for why.

PR #308, opened 2019-09-12
Render android

ceyusa:
Add GL render for android target, which is a unix type with only EGL.


ceyusa:
removed the GL context activation

ceyusa:
Can we merge this for now? @ferjm, @asajeffrey ?

ferjm:
Yes, I think we can merge as soon as CI is happy.

asajeffrey:
Fine by me. it would be nice if we could figure out a way to use features rather than gating on `target_os`, to reduce the amount of code duplication, but that's an orthogonal issue.

PR #310, opened 2019-09-16
Expose methods to suspend and resume all media actions

ferjm:
This is required for fixing https://github.com/servo/servo/issues/22763 and https://github.com/servo/servo/issues/21989

ServoMedia shuts down AudioContexts and Players when these are dropped, but because Servo keeps a bfcache, DOM objects are not always released immediately after navigation. So we need a way to manually suspend playback.

A similar issue occurs with `getUserMedia`, where we keep input streams active after navigation. I'll fix it in a follow-up.

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -293,6 +293,26 @@ impl Backend for GStreamerBackend {
             }
         }
     }
+
+    fn resume(&self, id: &ClientContextId) {
+        let mut instances = self.instances.lock().unwrap();
+        match instances.get_mut(id) {
+            Some(vec) => vec.retain(|(_, weak)| {
+                if let Some(mutex) = weak.upgrade() {
+                    let instance = mutex.lock().unwrap();
+                    if instance.resume().is_err() {
+                        warn!("Could not resume media instance");
+                    }
+                    true
+                } else {
+                    false
+                }
+            }),
+            None => {
+                warn!("Trying to resume an unknown client context");
+            }
+        }
+    }
 }
 
 impl AudioBackend for GStreamerBackend {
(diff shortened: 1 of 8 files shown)


backends/gstreamer/lib.rs
```diff
@@ -273,6 +273,26 @@ impl Backend for GStreamerBackend {
             }
         }
     }
+
+    fn suspend(&self, id: &ClientContextId) {
+        let mut instances = self.instances.lock().unwrap();
+        match instances.get_mut(id) {
+            Some(vec) => vec.retain(|(_, weak)| {
```
jdm: Should we abstract this operation (iterating over all known instances to perform an operation while dropping any dead ones) into a helper function that takes a closure argument, since it's used in two different places now?

ferjm: Done

ferjm:
r? @jdm 

I have added an additional `resume` method required to resume playback when navigating through the browsing history.

backends/gstreamer/lib.rs
```diff
@@ -139,6 +140,29 @@ impl GStreamerBackend {
             backend_chan: Arc::new(Mutex::new(backend_chan)),
         }))
     }
+
+    fn media_instance_action(
+        &self,
+        id: &ClientContextId,
+        cb: Rc<dyn Fn(Arc<Mutex<dyn MediaInstance>>) -> Result<(), ()>>,
```
jdm: It's not clear why the Rc is needed here - what if we use `&dyn Fn(...) -> ...` instead, and don't add `move` to each closure? Also, why not call `lock().unwrap()` in this method and pass `&mut dyn MediaInstance` to the callback instead?

PR #311, opened 2019-09-17
Sync each frame on the renderer's GL context not the app context

asajeffrey:
Sync the GL video frames on the renderer's GL context rather than the app's GL context.

### DIFF
--- a/backends/gstreamer/render-android/lib.rs
+++ b/backends/gstreamer/render-android/lib.rs
@@ -159,7 +159,13 @@ impl Render for RenderAndroid {
 
         if self.gst_context.lock().unwrap().is_some() {
             if let Some(sync_meta) = frame.buffer().get_meta::<gst_gl::GLSyncMeta>() {
-                sync_meta.wait(&self.app_context);
+                // This should possibly be
+                // sync_meta.wait(&self.app_context);
+                // since we want the main app thread to sync it's GPU pipeline too,
+                // but the main thread and the app context aren't managed by gstreamer,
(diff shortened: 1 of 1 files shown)


ceyusa:
Though call.

The gl wait/sync operations in android-render where taken from [my playground branch](https://github.com/ceyusa/media/commit/91891eb3e6cdd1c42a34b66d9ec1714d21031fc3) by accident (sorry :( ) when I added the render-android crate.

If you check the current render-unix crate, right now there are no sync/wait calls.

As @asajeffrey [mentioned first](https://mozilla.logbot.info/servo-magicleap/20190917#c16622734), the sync point, as far as I know -and as the code is right now-, should be the application context. Also, as @asajeffrey (mentioned next](https://mozilla.logbot.info/servo-magicleap/20190917#c16622782), and that is wrong in render-android, the wait call should be don in the render thread.

I would like to know (as I don't have the hardware to test it) if it is posible to remove the sync/wait calls and see if it still works.

With this patch, as far as I understand, basically we will doing a noop.

I'm sorry for all this hassle :(

asajeffrey:
My memory is that without the sync's the video doesn't show, presumably the sync's are doing some buffer swapping under the hood. I'll double check that when I'm in the office later today.

asajeffrey:
OK, double-checked and my memory is correct, without the sync's the video does not display.

asajeffrey:
r+ on IRC: https://mozilla.logbot.info/servo/20190918#c16626257

@bors-servo r=ceyusa

PR #312, opened 2019-09-20
Re-export gstreamer log related functions

ferjm:
This will help with https://github.com/servo/servo/issues/23700

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -71,6 +71,13 @@ use std::sync::{Arc, Mutex, Weak};
 use std::thread;
 use std::vec::Vec;
 
+// Re-export gstreamer log related functionality.
+// XXX(ferjm) ideally we should have an abstraction for this at the servo-media level.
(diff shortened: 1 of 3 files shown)


backends/gstreamer/lib.rs
```diff
@@ -71,6 +71,13 @@ use std::sync::{Arc, Mutex, Weak};
 use std::thread;
 use std::vec::Vec;
 
+// Re-export gstreamer log related functionality.
+// XXX(ferjm) ideally we should have an abstraction for this at the servo-media level.
+pub use gst::{
+    debug_add_log_function, debug_remove_default_log_function, DebugCategory, DebugLevel,
+    DebugLogFunction,
+};
```
ceyusa: You're exposing internal GStreamer API, breaking the backend independence.

jdm:
Rather than exposing some kind of logging API from this crate, I am inclined to have any embedder that needs special logging handling to initialize gstreamer and set it up themselves.

PR #314, opened 2019-09-27
Try first wayland and x11 displays before egl

ceyusa:
If EGLDisplay is used in a Wayland environment is incorrect, since
gstreamer will instanciate a dummy window not a wayland surface.


ceyusa:
@ferjm r?

PR #323, opened 2019-11-21
Support audio decoders for inputs with unusual numbers of channels.

jdm:
These changes address the linking failure present in https://github.com/servo/servo/issues/24803. The mix-matrix it is the part I'm least confident about; the channel mask for the decodebin is reported as 0x000000000000003f and the one for the audioconvert element is 0x0000000000000000, and https://gstreamer.freedesktop.org/documentation/audioconvert/index.html?gi-language=c mentions "A mix matrix can be passed to audioconvert, that will govern the remapping of input to output channels." along with "If an empty mix matrix is specified, a (potentially truncated) identity matrix will be generated." and has examples of using an empty one.

### DIFF
--- a/backends/gstreamer/audio_decoder.rs
+++ b/backends/gstreamer/audio_decoder.rs
@@ -147,6 +147,9 @@ impl AudioDecoder for GStreamerAudioDecoder {
                 let convert = gst::ElementFactory::make("audioconvert", None).ok_or(
                     AudioDecoderError::Backend("audioconvert creation failed".to_owned()),
                 )?;
+                convert
+                    .set_property("mix-matrix", &gst::Array::new(&[]).to_value())
+                    .expect("mix-matrix property didn't work");
                 let resample = gst::ElementFactory::make("audioresample", None).ok_or(
                     AudioDecoderError::Backend("audioresample creation failed".to_owned()),
                 )?;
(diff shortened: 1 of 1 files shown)


jdm:
@sdroege You might have useful input on whether this is an appropriate solution.

jdm:
Review ping @ferjm :)

ferjm:
Sorry, I totally forgot about this one.

LGTM, but I'd love to hear from @sdroege as well :)

@bors-servo r+

sdroege:
Sorry for seeing it just now. This change is not correct, but from this here or the connected issue I don't get enough information about what the problem actually is. Can you get me a full debug log of the problem with `GST_DEBUG=6`? Not just parts of it. Then I can take a look.

jdm:
@sdroege http://joshmatthews.net/gst

sdroege:
What is this `capsfilter0` on which you set `"audio/x-raw, format=(string)F32LE, layout=(string)interleaved, rate=(int)44100, channels=(int)6, channel-mask=(bitmask)0x0000000000000000"` as caps? That's the reason why there is a `0x0` channel-mask somewhere and why everything fails later.

sdroege:
Ah from `audio_decoder.rs`. You're doing there

```rust
                let mut audio_info_builder = gst_audio::AudioInfo::new(
                    gst_audio::AUDIO_FORMAT_F32,
                    options.sample_rate as u32,
                    channels,
                );
                if let Some(positions) = sample_audio_info.positions() {
                    audio_info_builder = audio_info_builder.positions(positions);
                }
```

For >2 channels that comes up with a channel-mask of `0x0` then because you have to explicitly give a channel-mask for complete caps in that case. If you only want filter caps to force a specific number of channels, etc here you should create incomplete caps and not use audioinfo.

```rust
let caps = gst::Caps::builder("audio/x-raw")
    .field("format", &gst_audio::AUDIO_FORMAT_F32.as_str())
    .field("rate", &(options.sample_rate as i32))
    .field("channels", &channels)
    .build();
if let Some(positions) = sample_audio_info.positions() {
    caps.set_simple(&[("channel-mask", &gst::Bitmask(gst_audio::AudioChannelPosition::positions_to_mask(positions).unwrap()))]);
}
```

Or something similar to that.

Issue #205, opened 2019-02-15
Implement WaveShaper node


jdm:
Spec: https://webaudio.github.io/web-audio-api/#waveshapernode

Reference processing code: https://searchfox.org/mozilla-central/source/dom/media/webaudio/WaveShaperNode.cpp

jdm:
For ease of implementation, don't both supporting oversampling at first.

khodzha:
I'd like to take a shot at oversampling, but how should it be implemented? Is adapting speex resampler code ok? Or maybe its better to use some existing lib?

jdm:
Adapting the speex code seems like a reasonable choice.

khodzha:
Ok, thanks, I'm not sure in my ability but I'll try.

jdm:
Yeah, https://github.com/xiph/speexdsp/blob/master/libspeexdsp/resample.c looks pretty complicated  :( It _might_ be worth trying out https://github.com/suhr/resample/blob/master/src/lib.rs?

khodzha:
@jdm i arrived at some 2x upsampling implementation but im not sure how to structure `Block` with 2x `buffer` len.

Right now block's buffer is hardcoded to store `channels*FRAMES_PER_BLOCK_USIZE` (so `channels*128`) elements.
I suppose I could store `Tick(256)` inside an upsampled block; but since [this comment](https://github.com/servo/media/blob/2ea0e52baf1ff10f94be6f4c507287f326add445/audio/block.rs#L15) says `Tick` is `time taken for a single frame` using `Tick(256)` looks wrong to me, since this is just 2x more samples per same Tick(128).

So idk how to proceed here, should I store something like frequency multiplier in a block?

Manishearth:
@khodzha don't, use 2x or 4x the blocks before wave shaping. Firefox does something similar. WaveShaper is supposed to downsample after upsampling so it's fine if a block is temporarily representing only half or quarter of a frame.

khodzha:
To be honest I'm kind of stuck wondering if my oversampling is useful at all, probably I'm missing something? I dont have any experience with digital signal processing

Right now my 2x oversampling implementation upsamples like in https://github.com/suhr/resample/blob/master/src/lib.rs using method described here: https://ccrma.stanford.edu/~jos/resample/
(basically every even output sample is a copy of corresponding input sample; every odd sample is linear combination of input samples weighted with kaiser-windowed sincs)

[...]

PR #332, opened 2020-02-02
2x and 4x oversampling for WaveShaperNode

khodzha:
Issue #205 
Finally managed to defeat the noise I mentioned in the issue 

I'm not sure if ditching FrameIterator and operating on raw vecs was a good thing :man_shrugging: 

Also spec mentions tail-time but I'm not sure how to handle that

### DIFF
--- a/audio/Cargo.toml
+++ b/audio/Cargo.toml
@@ -17,6 +17,7 @@ servo_media_derive = { path = "../servo-media-derive" }
 servo-media-player = { path = "../player" }
 servo-media-traits = { path = "../traits" }
 smallvec = "0.6.1"
+speexdsp-resampler = "0.1.0"
 
 [dependencies.petgraph]
 version = "0.4.12"
--- a/audio/wave_shaper_node.rs
+++ b/audio/wave_shaper_node.rs
@@ -1,11 +1,31 @@
-use block::Chunk;
+use block::{Chunk, FRAMES_PER_BLOCK_USIZE};
 use node::{AudioNodeEngine, AudioNodeType, BlockInfo, ChannelInfo};
+use speexdsp_resampler::State as SpeexResamplerState;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum OverSampleType {
     None,
     Double,
     Quadruple,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+enum TailtimeBlocks {
+    Zero,
+    One,
+    Two,
+}
+
+const OVERSAMPLING_QUALITY: usize = 0;
+
+impl OverSampleType {
+    fn value(&self) -> usize {
+        match self {
+            OverSampleType::None => 1,
+            OverSampleType::Double => 2,
+            OverSampleType::Quadruple => 4,
+        }
+    }
 }
 
 type WaveShaperCurve = Option<Vec<f32>>;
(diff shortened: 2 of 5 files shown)


audio/Cargo.toml
```diff
@@ -18,6 +18,10 @@ servo-media-player = { path = "../player" }
 servo-media-traits = { path = "../traits" }
 smallvec = "0.6.1"
 
+[dependencies.speexdsp]
+git = "https://github.com/rust-av/speexdsp-rs"
```
Manishearth: Not https://crates.io/crates/speexdsp ?

audio/wave_shaper_node.rs
```diff
@@ -116,3 +165,30 @@ impl AudioNodeEngine for WaveShaperNode {
 
     make_message_handler!(WaveShaperNode: handle_waveshaper_message);
 }
+
+impl WaveShaperNode {
+    fn apply_curve(buf: &mut [f32], curve: &Vec<f32>) {
+        buf.iter_mut().for_each(|sample| {
+            let len = curve.len();
+            let curve_index: f32 = ((len - 1) as f32) * (*sample + 1.) / 2.;
```
Manishearth: worth  moving `(len - 1) / 2` and `len` out of the closure. Probably will be done by the compiler anyway, but doesn't hurt.

audio/Cargo.toml
Manishearth: Ah, it's not released yet.

Manishearth: Looks like they're willing to make one soon! https://github.com/rust-av/speexdsp-rs/issues/10

audio/wave_shaper_node.rs
khodzha: agreed, I'll fix it up tomorrow

lu-zero:
I prepared a crate with the stand alone resampler, if you could please confirm that the API exposed is fine for you I'll publish it.

Manishearth:
@khodzha ^^

khodzha:
Updated PR to use this crate, seems ok to me :man_shrugging:

lu-zero:
https://crates.io/crates/speexdsp-resampler published !

audio/wave_shaper_node.rs
```diff
@@ -79,33 +93,68 @@ impl AudioNodeEngine for WaveShaperNode {
         AudioNodeType::WaveShaperNode
     }
 
-    fn process(&mut self, mut inputs: Chunk, _info: &BlockInfo) -> Chunk {
+    fn process(&mut self, mut inputs: Chunk, info: &BlockInfo) -> Chunk {
         debug_assert!(inputs.len() == 1);
 
         if inputs.blocks[0].is_silence() {
```
khodzha: @Manishearth could you confirm whether tail-time here could be implemented like this:
if we hit two silent blocks in row after non-silent block, we `explicit_silence()` them and run them through the waveshaping so all the tail-time from resampler would be spitted out

Manishearth: That sounds correct.

khodzha:
Added tail-time handling and properly handled the case when waveshaped silence is nonsilent

Manishearth:
@bors-servo r+

thanks!

Issue #306, opened 2019-09-11
ServoMedia::get is unsound


nox:
The initialization of `ServoMedia` is very much unsound and should just be using `lazy_static`.

This is `INSTANCE`, note how it is initialized as a raw null pointer (that should be using `std::ptr::null_mut()` btw):

https://github.com/servo/media/blob/a70f02482d29472c5566e16ffa934fda909443bb/servo-media/lib.rs#L25

Those are the three methods handling initialization:

https://github.com/servo/media/blob/a70f02482d29472c5566e16ffa934fda909443bb/servo-media/lib.rs#L68-L90

Note how calling `ServoMedia::get()` before `ServoMedia::init()` or `ServoMedia::init_with_backend(backend)` will dereference a null pointer.

Please use [`lazy_static`](https://docs.rs/lazy_static) instead of manually using `Once` and a raw pointer in a static.

nox:
The raw pointer and `Once` value was [previously](https://github.com/servo/media/commit/6e2ce72ba4422229d9756118c7b37d5d3d830cb2) soundly used but it was [broken](https://github.com/servo/media/commit/a94a3b57ae5dbee9481fb98a1e2dd12e5303e8db) when the method `init` was introduced.

khodzha:
How does one pass `Box<dyn Backend>` into `lazy_static` block?

Manishearth:
#335

PR #335, opened 2020-04-13
return err if ServoMedia wasnt initialized

khodzha:
im not sure how to do this with lazy_static since init requires type parameter
but at least this addresses null pointer dereference

issue at hand: #306


Manishearth:
Thanks!

PR #337, opened 2020-05-05
IIR filter node implementation

khodzha:


khodzha:
I'm not sure whether `Arc` is overkill but i failed to compile it with raw references

also spec says there must be [`getFrequencyResponse`](https://webaudio.github.io/web-audio-api/#IIRFilterNode-methods) method which *synchronously* calculates the frequency response, and I'm not sure whether `handle_message` is a proper way to do this? @ferjm

khodzha:
also insert(0) isnt very bad, should improve that

khodzha:
i still dont know about proper `getFrequencyResponse` implementation but i added a method so at least there is something to fix

audio/iir_filter_node.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        let output = (inputs_sum - outputs_sum) / self.feedback[0];
+
+        if output.is_nan() {
+            // Per spec:
+            // Note: The UA may produce a warning to notify the user that NaN values have occurred in the filter state.
+            // This is usually indicative of an unstable filter.
+            //
+            // But idk how to produce warnings
+            panic!("NaN in IIRFilter state");
```
ferjm, line 62: A panic here may be too much. Let's `warn!` instead.

Unfortunately, AFAICT we have no way to progress this warning up to the DOM code at the moment.

audio/iir_filter_node.rs
```diff
@@ -0,0 +1,177 @@
...
+            panic!("NotSupportedError: feedback must have at least one coeff");
+        }
+
+        if options.feedback.len() > MAX_COEFFS {
+            panic!("NotSupportedError: feedback max length is {}", MAX_COEFFS);
+        }
+
+        if options.feedback[0] == 0.0 {
+            panic!("InvalidStateError: first feedback coeff must not be zero");
+        }
```
ferjm, line 110: In theory we should never get to these `panic`s as these are DOM level checks. So we usually do not double check these conditions in servo-media. I think it would be better to add `debug_assert`s instead of `panic`s.

audio/iir_filter_node.rs
```diff
@@ -0,0 +1,177 @@
...
+            filters: vec![filter; channel_info.count as usize],
+            channel_info,
+            feedforward,
+            feedback,
+        }
+    }
+
+    pub fn get_frequency_response(&self, frequency_hz: &[f32], mag_response: &mut [f32], phase_response: &mut [f32]) {
+        if frequency_hz.len() != mag_response.len() || frequency_hz.len() != phase_response.len() {
+            // TODO: InvalidAccessError
```
ferjm, line 126: Remove this TODO comment, please

audio/iir_filter_node.rs
```diff
@@ -0,0 +1,177 @@
...
+            panic!("InvalidAccessError: getFrequencyResponse params are of different length");
```
ferjm, line 127: Same as above. s/panic/debug_assert

khodzha:
addressed the issues + implemented `IIRFilterNodeMessage::GetFrequencyResponse`

ferjm:
Nice. Thanks!

@bors-servo r+

Manishearth:
I would rather not introduce further synchronous messaging traffic; it is very easy to deadlock here. The script crate should not be blocking on the audio crate.

The way to expose this method would be to expose a `get_frequency_response()` utility function that uses the IIR filter params stored on the script-side, along with its parameters, to compute the frequency response.

khodzha:
by free you mean
```
pub fn get_frequency_response(
        feedforward: &[f32],
        feedback: &[f32],
        frequency_hz: &[f32],
        mag_response: &mut [f32],
        phase_response: &mut [f32],
    ) -> {}
```
without any `self` reference?

Manishearth:
Yeah, basically.

PR #339, opened 2020-05-05
set AudioParams as KRate in AudioBufferSourceNode

khodzha:


audio/listener.rs
```diff
@@ -31,15 +31,15 @@ impl AudioListenerNode {
...
-            position_x: Param::new(0.),
-            position_y: Param::new(0.),
-            position_z: Param::new(0.),
-            forward_x: Param::new(0.),
-            forward_y: Param::new(0.),
-            forward_z: Param::new(-1.),
-            up_x: Param::new(0.),
-            up_y: Param::new(1.),
-            up_z: Param::new(0.),
+            position_x: Param::new_krate(0.),
```
Manishearth, line 34: This doesn't seem correct. Where in the spec does this happen?

khodzha: https://webaudio.github.io/web-audio-api/#AudioParam-attributes

> If the panningModel is "HRTF", the setting of the automationRate for any AudioParam of the PannerNode is ignored. Likewise, the setting of the automationRate for any AudioParam of the AudioListener is ignored. In this case, the AudioParam behaves as if the automationRate were set to "k-rate".

Manishearth: This should happen in the panning implementation, the panner node is ignoring the automationRate, the listener still has the automation rate

khodzha:
i removed `new_krate` change from `AudioListener`
i think it should still be krate for `BufferSourceNode` and (not implemented) `DynamicCompressionNode`

PR #340, opened 2020-05-05
added AudioRenderThreadMsg::GetNodeType

khodzha:
Some `AudioParams` for some nodes arent allowed to change automation rate
to `ARate`.
To check for this in servo in `SetAutomationRate` `AudioNodeType` must be
exposed via `AudioContext` method.

This is preliminary step to implement these checks in [SetAutomationRate](https://github.com/servo/servo/blob/3ea6d87bcc37167464e856949a4b9b77d0e9318a/components/script/dom/audioparam.rs#L108)

### DIFF
--- a/audio/node.rs
+++ b/audio/node.rs
@@ -37,7 +37,7 @@ pub enum AudioNodeInit {
 }
 
 /// Type of AudioNodeEngine.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Eq)]
 pub enum AudioNodeType {
     /// Not a constructable node
     AudioListenerNode,


Manishearth:
We should definitely not be using the messaging system for this. I don't see why this is necessary, as I said this needs to be handled in the DOM implementation.

khodzha:
I'm probably missing something but how can one access node type if AudioParam only has NodeId?

https://github.com/servo/servo/blob/3ea6d87bcc37167464e856949a4b9b77d0e9318a/components/script/dom/audioparam.rs#L23-L32

Manishearth:
Stash a reference to the node, or the node type, on the param at creation

khodzha:
@Manishearth Sorry i suppose it must be a reference to a node since for `PannerNode` we need to query `panning_model`
But idk how to save it in `AudioParam` since every node in servo/script is its own type?  so idk which type should i cast all these nodes to

Manishearth:
They subclass AudioNode.

khodzha:
i'll open new PR

PR #341, opened 2020-05-07
implemented Eq on AudioNodeType

khodzha:
since its needed to compare AudioNodeTypes on AutomationRate change

### DIFF
--- a/audio/node.rs
+++ b/audio/node.rs
@@ -37,7 +37,7 @@ pub enum AudioNodeInit {
 }
 
 /// Type of AudioNodeEngine.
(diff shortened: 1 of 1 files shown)


audio/node.rs
```diff
@@ -37,7 +37,7 @@ pub enum AudioNodeInit {
 }
 
 /// Type of AudioNodeEngine.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Eq)]
```
jdm, line 40: This needs PartialEq as well.

khodzha: yeah :rofl:

PR #346, opened 2020-05-25
Add required rand dependency to webrtc example.

jdm:


sdroege:
Unrelated to this PR, but the example looks quite dated. Might be worth basing a new one on https://github.com/centricular/gstwebrtc-demos/tree/master/sendrecv/gst-rust, with `async`/`await` goodness

sdroege:
Unrelated to this PR, but the example looks quite dated. Might be worth basing a new one on https://github.com/centricular/gstwebrtc-demos/tree/master/sendrecv/gst-rust, with `async`/`await` goodness

PR #348, opened 2020-05-29
Add MediaStreamDestinationNode

### LINKED ISSUE #342: Implement MediaStreamAudioDestinationNode
Needed for Hubs: https://github.com/servo/servo/issues/26097

This is a node that outputs to a MediaStream.

We'll need to create a node with an internal appsrc-audioconvert-(mediastream proxysink) pipeline.

A tricky part here is that currently servo-media-audio graph traversal cares only about things reachable by the normal destination node. We may need to introduce a "parent destination" node in the graph traversal that all destination nodes inherit from.

Manishearth:
One suboptimal part is that currently you need to block on a receiver to get the media stream id.

There are a couple better ways to do that:

 - Pass in the AudioSink from the client
 - Tweak the streams api to allow creating a media stream that has no input yet, and connect it up to

Fixes #342 

Thoughts?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2143,6 +2143,7 @@ dependencies = [
  "serde 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.80 (registry+https://github.com/rust-lang/crates.io-index)",
  "servo-media-player 0.1.0",
+ "servo-media-streams 0.1.0",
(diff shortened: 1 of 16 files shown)


ferjm:
LGTM. It may be worth creating a follow up issue to explore the alternatives to blocking on stream ID generation that you mentioned.

Thanks!

@bors-servo r+

PR #349, opened 2020-05-31
fixed BufferSourceNode interpolation when pos is zero

khodzha:
fixes https://github.com/servo/servo/issues/26663


ferjm:
Thanks!

@bors-servo r+

PR #350, opened 2020-05-31
WebRTC data channels backend

ferjm:
This is not entirely spec compliant, but it contains the basic stuff to make data channels work. It allows sending only strings for now.

### DIFF
--- a/backends/gstreamer/datachannel.rs
+++ b/backends/gstreamer/datachannel.rs
@@ -2,7 +2,7 @@ use glib::{ObjectExt, Value};
 use servo_media_webrtc::thread::InternalEvent;
 use servo_media_webrtc::WebRtcController as WebRtcThread;
 use servo_media_webrtc::{
-    DataChannelEvent, DataChannelId, DataChannelInit, WebRtcError, WebRtcResult,
+    DataChannelEvent, DataChannelId, DataChannelInit, DataChannelMessage, WebRtcError, WebRtcResult,
 };
 use std::sync::Mutex;
 
@@ -14,11 +14,18 @@ use std::sync::Mutex;
 struct DataChannel(glib::Object);
 
 impl DataChannel {
-    pub fn send(&self, message: &str) -> WebRtcResult {
-        self.0
-            .emit("send-string", &[&Value::from(message)])
-            .map(|_| ())
-            .map_err(|e| WebRtcError::Backend(e.to_string()))
+    pub fn send(&self, message: &DataChannelMessage) -> WebRtcResult {
+        match message {
+            DataChannelMessage::Text(message) => {
+                self.0.emit("send-string", &[&Value::from(&message)])
+            }
+            DataChannelMessage::Binary(message) => {
+                let bytes = glib::Bytes::from(message);
+                self.0.emit("send-data", &[&Value::from(&bytes)])
+            }
+        }
+        .map(|_| ())
+        .map_err(|e| WebRtcError::Backend(e.to_string()))
     }
 
     pub fn close(&self) -> WebRtcResult {
@@ -143,7 +150,7 @@ impl GStreamerWebRtcDataChannel {
         })
     }
 
-    pub fn send(&self, message: &str) -> WebRtcResult {
+    pub fn send(&self, message: &DataChannelMessage) -> WebRtcResult {
         self.channel.send(message)
     }
 
--- a/backends/gstreamer/datachannel.rs
+++ b/backends/gstreamer/datachannel.rs
@@ -1,43 +1,38 @@
 use boxfnonce::SendBoxFnOnce;
 use glib::{ObjectExt, Value};
-use servo_media_webrtc::thread::InternalEvent;
-use servo_media_webrtc::WebRtcController as WebRtcThread;
 use servo_media_webrtc::{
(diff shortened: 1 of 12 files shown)


backends/gstreamer/datachannel.rs
```diff
@@ -0,0 +1,144 @@
...
+                    .expect("Invalid data channel message")
+                {
+                    callbacks_.lock().unwrap().message(message);
+                }
+                None
+            })
+            .map_err(|e| e.to_string())?;
+
+        let channel = InnerDataChannel {
+            channel: glib::SendWeakRef::from(channel.downgrade()),
```
sdroege, line 105: This is going to explode sooner or later. The datachannel object is thread-safe, but until https://gitlab.freedesktop.org/gstreamer/gst-plugins-bad/-/issues/1168 is solved you'll have to put your own wrapper struct around it on which `Send`/`Sync` are implemented.

backends/gstreamer/datachannel.rs
```diff
@@ -129,15 +111,11 @@ impl WebRtcDataChannelBackend for GStreamerWebRtcDataChannel {
...
-        // was created on the webrtc controller's thread.
-        self.webrtc_thread
-            .internal_event(InternalEvent::SendDataChannelMessage(
-                Box::new(self.channel.clone()),
-                message.to_owned(),
-            ));
-        Ok(())
+        self.channel
+            .0
+            .emit("send-string", &[&Value::from(message)])
```
sdroege, line 116: While at it, you could also hide the signal emission things in functions inside `impl DataChannel` :)

backends/gstreamer/datachannel.rs
```diff
@@ -0,0 +1,109 @@
...
+            })
+            .map_err(|e| e.to_string())?;
+
+        let channel = glib::SendWeakRef::from(channel.downgrade());
+        Ok(Self { channel, callbacks })
+    }
+}
+
+impl WebRtcDataChannel for GStreamerWebRtcDataChannel {
+    fn set_on_open(&self, cb: SendBoxFnOnce<'static, ()>) {
```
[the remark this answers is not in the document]
Manishearth: `send()` can be called by the controller instead

webrtc/lib.rs
```diff
@@ -50,11 +52,10 @@ pub trait WebRtcControllerBackend: Send {
     fn create_offer(&mut self, cb: SendBoxFnOnce<'static, (SessionDescription,)>) -> WebRtcResult;
     fn create_answer(&mut self, cb: SendBoxFnOnce<'static, (SessionDescription,)>) -> WebRtcResult;
     fn add_stream(&mut self, stream: &MediaStreamId) -> WebRtcResult;
-    fn create_data_channel(
-        &mut self,
-        init: &WebRtcDataChannelInit,
-        channel: Sender<Box<dyn WebRtcDataChannelBackend>>,
-    ) -> WebRtcResult;
+
+    fn create_data_channel(&mut self, id: &DataChannelId, init: &DataChannelInit) -> WebRtcResult;
```
Manishearth, line 56: the controller probably can manage ids itself and return an id here.

backends/gstreamer/datachannel.rs
```diff
@@ -45,83 +57,96 @@ impl GStreamerWebRtcDataChannel {
...
+                            DataChannelEvent::Error(WebRtcError::Backend(error.to_string())),
+                        ));
                 }
                 None
             })
             .map_err(|e| e.to_string())?;
 
-        let callbacks_ = callbacks.clone();
+        let id_ = id.clone();
+        let thread_ = Mutex::new(thread.clone());
```
Manishearth, line 119: the outer mutex doesn't seem necessary?

Manishearth: @sdroege should the `on-message-string` and `on-error` callbacks require Sync here? I recall we fixed some of these cases earlier: https://github.com/sdroege/gstreamer-rs/issues/154

webrtc/datachannel.rs
```diff
@@ -0,0 +1,83 @@
+use crate::{WebRtcError, WebRtcResult};
+
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use uuid::Uuid;
+
+lazy_static! {
```
Manishearth, line 7: I think these ids should just be numbers and individual webrtc implementations should produce and consume them

We shouldn't have a `DataChannelBackend` trait at all, all of this should be internal to the webrtc impl

examples/simple_webrtc.rs
```diff
@@ -214,10 +198,30 @@ impl WebRtcSignaller for Signaller {
...
-        }));
+    fn on_data_channel_event(
+        &self,
+        id: DataChannelId,
+        event: DataChannelEvent,
+        controller: &WebRtcController,
+    ) {
+        match event {
+            DataChannelEvent::NewChannel => {
+                println!("New channel {:?}", id);
```
Manishearth, line 209: probably should be removed or turned into `debug!()`

backends/gstreamer/datachannel.rs
ferjm: `rustc` complains if I don't add the outer mutex

Manishearth: Yeah sorry this is the Sync requirement that I think needs to be removed upstream. No worries for now

PR #357, opened 2020-06-15
Separate out encoding of media streams

### LINKED ISSUE #354: Deal with the encoded/unencoded distinction in MediaStream
Currently, GStreamerMediaStream can contain either encoded or raw audio/video. The current situation is:

 - streams produced by webrtc are raw
 - streams _consumed_ by webrtc must be encoded
 - media capture starts out raw but we produce it encoded so that it can be fed to webrtc
 - player consumes streams encoded

IMO we should make it so that all streams are raw by default, but you can choose to stick in an encoder pipeline if you wish.

Manishearth:
This fixes https://github.com/servo/media/issues/354 by normalizing all streams to contain raw audio. Instead, webrtc is given a way to obtain an encoded adapter for a given media stream.

Fixes #354

### DIFF
--- a/backends/gstreamer/media_stream.rs
+++ b/backends/gstreamer/media_stream.rs
@@ -87,13 +87,6 @@ impl GStreamerMediaStream {
         }
     }
 
-    pub fn insert_capsfilter(&mut self) {
-        assert!(self.pipeline.is_none());
-        let capsfilter = gst::ElementFactory::make("capsfilter", None).unwrap();
-        capsfilter.set_property("caps", self.caps()).unwrap();
-        self.elements.push(capsfilter);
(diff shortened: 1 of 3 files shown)


backends/gstreamer/media_stream.rs
```diff
@@ -102,6 +95,7 @@ impl GStreamerMediaStream {
         assert!(self.pipeline.is_none());
         let elements: Vec<_> = self.elements.iter().collect();
         pipeline.add_many(&elements[..]).unwrap();
+        println!("LINKING {:?}", &elements);
```
ferjm, line 98: Use `debug!` or remove this line, please.

PR #364, opened 2020-06-24
Make sure that we send NewChannel before any other event

ferjm:


ferjm:
r? @jdm 

I think this fixes the issue that you mentioned [here](https://github.com/servo/servo/pull/26752#discussion_r444472272)

backends/gstreamer/webrtc.rs
```diff
@@ -76,6 +76,7 @@ pub struct GStreamerWebRtcController {
     remote_offer_generation: u32,
     _main_loop: glib::MainLoop,
     data_channels: Arc<Mutex<HashMap<DataChannelId, GStreamerWebRtcDataChannel>>>,
+    data_channels_event_queue: Arc<Mutex<HashMap<DataChannelId, Vec<DataChannelEvent>>>>,
```
Manishearth, line 79: A concurrent queue is just a sender/receiver pair

jdm, line 79: I get a little bit nervous about races when there are multiple independent locks for what is essentially the same data. Is there any way to do something like:
```rust
enum DataChannelState {
    Buffered(Vec<DataChannelEvent>),
    Created(GStreamerWebRtcDataChannel),
}
```
And then only have a single hashmap?

Manishearth: I think we can have a single hashmap adn then just a Sender/Receiver pair, the vec is being used as a queue.

But the enum works too

backends/gstreamer/webrtc.rs
```diff
@@ -168,7 +179,10 @@ impl WebRtcControllerBackend for GStreamerWebRtcController {
         message: &DataChannelMessage,
     ) -> WebRtcResult {
         match self.data_channels.lock().unwrap().get(id) {
-            Some(ref channel) => channel.send(message),
+            Some(ref channel) => match channel {
+                DataChannelEventTarget::Created(ref channel) => channel.send(message),
+                _ => Ok(()),
```
jdm, line 184: It seems a bit worrying that we silently ignore messages here. We could get here by calling create_data_channel then immediately calling send_data_channel_message - should we return an error instead?

backends/gstreamer/webrtc.rs
```diff
@@ -571,23 +593,33 @@ impl GStreamerWebRtcController {
                 let thread_ = thread.lock().unwrap().clone();
                 match GStreamerWebRtcDataChannel::from(&id, channel, &thread_) {
                     Ok(channel) => {
+                        {
+                            let mut data_channels = data_channels.lock().unwrap();
+                            if let Some(ref mut channel) = data_channels.get_mut(&id) {
+                                match channel {
+                                    DataChannelEventTarget::Buffered(ref mut events) => {
+                                        for event in events.drain(0..) {
+                                            thread_.internal_event(
```
jdm, line 602: Is there any chance that we could have a buffered close event? If so, we will end up registering a channel after the close.

backends/gstreamer/webrtc.rs
jdm: Nevermind, I was confused.

jdm:
r=me when you're satisfied with the answer to my question.

ferjm:
This was merged already  6aa53e1

PR #367, opened 2020-07-01
Move media stream setter later

Manishearth:
Turns out the API requires deferred initialization


Manishearth:
No, never mind, I read this wrong :)

PR #368, opened 2020-07-02
Device monitor

ferjm:
This is required for https://github.com/servo/servo/pull/27152/files

### DIFF
--- a/backends/gstreamer/device_monitor.rs
+++ b/backends/gstreamer/device_monitor.rs
@@ -1,0 +1,61 @@
+use atomic_refcell::AtomicRefCell;
+use gst::Caps;
+use gst::DeviceExt;
+use gst::DeviceMonitor as GstDeviceMonitor;
+use gst::DeviceMonitorExt;
+use gst::DeviceMonitorExtManual;
+
+use servo_media::{MediaDeviceInfo, MediaDeviceKind};
+
+pub struct GStreamerDeviceMonitor {
+    devices: AtomicRefCell<Option<Vec<MediaDeviceInfo>>>,
+}
+
+impl GStreamerDeviceMonitor {
+    pub fn new() -> GStreamerDeviceMonitor {
+        GStreamerDeviceMonitor {
+            devices: AtomicRefCell::new(None),
+        }
+    }
+
+    fn get_devices(&self) -> Result<Vec<MediaDeviceInfo>, ()> {
+        const AUDIO_SOURCE: &str = "Audio/Source";
+        const AUDIO_SINK: &str = "Audio/Sink";
+        const VIDEO_SOURCE: &str = "Video/Source";
+        let device_monitor = GstDeviceMonitor::new();
+        let audio_caps = Caps::new_simple("audio/x-raw", &[]);
+        device_monitor.add_filter(Some(AUDIO_SOURCE), Some(&audio_caps));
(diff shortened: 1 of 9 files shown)


backends/gstreamer/device_monitor.rs
```diff
@@ -0,0 +1,61 @@
...
+            .iter()
+            .map(|device| {
+                let display_name = device.get_display_name().as_str().to_owned();
+                MediaDeviceInfo {
+                    device_id: display_name.clone(),
+                    kind: match device.get_device_class().as_str() {
+                        AUDIO_SOURCE => MediaDeviceKind::AudioInput,
+                        AUDIO_SINK => MediaDeviceKind::AudioOutput,
+                        VIDEO_SOURCE => MediaDeviceKind::VideoInput,
+                        _ => MediaDeviceKind::__Unknown,
```
Manishearth, line 42: We should not have this variant, we should just filter_map these out

backends/gstreamer/device_monitor.rs
```diff
@@ -0,0 +1,61 @@
...
+use gst::Caps;
+use gst::DeviceExt;
+use gst::DeviceMonitor as GstDeviceMonitor;
+use gst::DeviceMonitorExt;
+use gst::DeviceMonitorExtManual;
+
+use servo_media::{MediaDeviceInfo, MediaDeviceKind};
+
+pub struct GStreamerDeviceMonitor {
+    devices: AtomicRefCell<Option<Vec<MediaDeviceInfo>>>,
```
Manishearth, line 11: is the atomic refcell necessary?

can we instead return a GSTDeviceMonitor and have `navigator.mediadevices` contain one via a trait object?

ferjm: Sorry, I can't see how returning a `GstDeviceMonitor` makes the atomic refcell go away. The atomic refcell is here just to allow caching the result of the `GStreamerDeviceMonitor.enumerate_devices` call. If you think it is preferable to do the caching on the `navigator.mediaDevices` side, we can move the cache there, but I don't see the need to return the extra trait object.

Manishearth: Yeah I want to do the caching on the mediadevices side.

The trait object lets us return a DeviceMonitor object that contains the cache (and will only be accessed from one thread), but it's also fine if we do the caching in servo.

Might need tweaks once we decide to support devicechange.

backends/gstreamer/lib.rs
```diff
@@ -1,4 +1,6 @@
 #![feature(nll)]
+
+extern crate atomic_refcell;
```
Manishearth, line 3: unnecessary import

backends/gstreamer/device_monitor.rs
```diff
@@ -0,0 +1,63 @@
...
+            })
+            .collect();
+        Ok(devices)
+    }
+}
+
+impl MediaDeviceMonitor for GStreamerDeviceMonitor {
+    fn enumerate_devices(&self) -> Result<Vec<MediaDeviceInfo>, ()> {
+        {
+            if let Some(ref devices) = *self.devices.borrow() {
```
Manishearth, line 55: we should probably file a followup about the devices changing

Manishearth: though it's not that important, actually, sicne we don't use this cache in servo

PR #372, opened 2020-07-14
[WIP] Allow consuming stream data from a client app

ferjm:
This is required to implement https://github.com/servo/servo/issues/26861

Still untested

### DIFF
--- a/backends/gstreamer/media_stream.rs
+++ b/backends/gstreamer/media_stream.rs
@@ -2,6 +2,7 @@ use super::BACKEND_BASE_TIME;
 use glib::prelude::*;
 use gst;
 use gst::prelude::*;
+use gst_app::AppSrc;
 use servo_media_streams::registry::{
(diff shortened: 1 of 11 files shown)


ferjm:
r? Manishearth

backends/gstreamer/media_stream.rs
```diff
@@ -186,14 +192,74 @@ impl GStreamerMediaStream {
         }
     }
 
-    pub fn create_video_from(source: gst::Element) -> MediaStreamId {
+    pub fn set_video_app_source(&mut self, source: &AppSrc) {
+        self.video_app_source = Some(source.clone());
+    }
+
+    pub fn create_video_from(source: gst::Element, size: Option<Size2D<u32>>) -> MediaStreamId {
```
Manishearth, line 199: We shouldn't be doing decoding within `create_video_from`, it's the responsibility of the client to do that (see webrtc, all media streams are raw, this was a deliberate choice and without it we will have problems)

PR #395, opened 2023-06-24
Update ipc-channel to 0.16

mukilan:
ipc-channel used in servo needs to bumped to 0.16.1 to complete https://github.com/servo/servo/issues/29467, but that requires the same version to be used by media crates as well.

FYI: I had accidentally pushed my change to master branch of this repo. I've now force pushed master to it's previous state. I've also made 'master' a protected branch. Please let me know if that needs to be changed or removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -385,19 +385,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.4.4"
+version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b153fe7cbef478c567df0f972e02e6d736db11affe43dfc9c56a9374d1adfb87"
-dependencies = [
- "crossbeam-utils 0.7.2",
- "maybe-uninit",
-]
-
-[[package]]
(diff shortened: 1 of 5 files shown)


jdm:
This looks fine, but apparently webrender no longer builds on nightly rust :(

mukilan:
@jdm looks like only the example code uses webrender, specifically v0.60.0 which is quite old and the offending code has been removed in servo/webrender@e8d2d7d6

I am able to build and run servo master with latest ipc-channel (plus your patch and another small change to account for the[ renaming of `drain_filter` ](https://github.com/rust-lang/rust/issues/43244) nightly feature flag.

Do you think we should first publish webrender 0.62.0 and update the failing examples or should we merge this PR now and fix the examples in a new issue?

cc @mrobinson as you know more about webrender

jdm:
Ah, if it's just example code than I'm less concerned. I'm ok with merging and fixing up the example separately.

mukilan:
I've created [this issue](https://github.com/servo/media/issues/396) to update the example with new webrender.

I'll merge this PR manually for now.

PR #399, opened 2023-08-24
Update ipc-channel

mrobinson:
This will allow upgrading rand in Servo.


Manishearth (approved):

mrobinson:
It seems that updating `rand` in the examples is a bit more complicated, so I'll put that aside for now and just land the upgrade to `ipc-channel`.

mrobinson:
Thanks for the review!

PR #401, opened 2023-09-06
Fix examples on CI

mrobinson:


.github/workflows/rust.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  GST_PLUGIN_FEATURE_RANK: pulsesink:NONE,alsasink:NONE,jacksink:NONE,fakesink:MAX
```
ceyusa, line 13: It's a good start but the envvar might grow with video sinks and so the test use them. 

overall, LGTM

ceyusa (approved):

mrobinson:
Thank you for the review!

mrobinson:
I guess this will need one more reviewer with write access to this repository to land.

jdm (approved):

PR #405, opened 2023-11-10
Initialize media in the background

mrobinson:
Move initialization of the media engine to a background thread. This
also replaced the complex INITIALIZER / INSTANCE combo with a single
synchronous `OnceCell`. Finally, instead of taking in the `Backend` as
an argument, take in a factory function. This allows the `Backend` to
also be initialized off the main thread.

This allows GStreamer plugin cache initialization to happen off the main
thread and removes unsafe code in initialization.

Co-authored-by: Rakhi Sharma <atbrakhi@igalia.com>

### DIFF
--- a/servo-media/Cargo.toml
+++ b/servo-media/Cargo.toml
@@ -7,6 +7,9 @@ authors = ["Manish Goregaokar <manishsmail@gmail.com>"]
 name = "servo_media"
 path = "lib.rs"
 
+[dependencies]
+once_cell = "1.18.0"
+
 [dependencies.servo-media-audio]
 path = "../audio"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  GST_PLUGIN_FEATURE_RANK: pulsesink:NONE,alsasink:NONE,jacksink:NONE,fakesink:MAX
+  GST_PLUGIN_FEATURE_RANK: openalsink:NONE,pulsesink:NONE,alsasink:NONE,jacksink:NONE,fakesink:MAX
```
atbrakhi, line 13: I am not sure if i understand this change. Is it relevant to this media issue?

mrobinson: Sorry, I'm trying to debug the CI failure which is unrelated to this change.

jdm (approved):
This is great!

atbrakhi (approved):

PR #406, opened 2023-11-13
Try to fix the CI by using a dummy jack device

mrobinson:
Instead of using GST_PLUGIN_FEATURE_RANK, which may not be supported by
our version of Gstreamer start a dummy jackd instance.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,7 +10,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  GST_PLUGIN_FEATURE_RANK: pulsesink:NONE,alsasink:NONE,jacksink:NONE,fakesink:MAX
 
 jobs:
   Build:
@@ -61,6 +60,10 @@ jobs:
           pulseaudio \
           pulseaudio-utils \
           webp
+    - name: Start virtual audio device
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        jackd -d dummy &
     - name: Install rust
       uses: actions-rs/toolchain@v1
       with:


.github/workflows/rust.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  GST_PLUGIN_FEATURE_RANK: pulsesink:NONE,alsasink:NONE,jacksink:NONE,fakesink:MAX
+  GST_PLUGIN_FEATURE_RANK: openalsink:NONE,openalsrc:NONE,pulsesink:NONE,alsasink:NONE,jackaudiosrc:NONE,jackaudiosink:NONE,fakesink:MAX
```
philn, line 13: Do you have `fakeaudiosink` ? Might be a better fit than `fakesink`

mrobinson: Good question. How do I know what sinks are available?

mrobinson: I don't see `fakeaudiosink` listed here for example: https://gstreamer.freedesktop.org/documentation/coreelements/fakesink.html?gi-language=c

philn: It's in -bad (staging area), https://gstreamer.freedesktop.org/documentation/debugutilsbad/fakeaudiosink.html?gi-language=c#fakeaudiosink-page

The best way to know if you have it is to check if its in the `gst-inspect-1.0` output.

rafaelcaricio:
@mrobinson maybe you want to open a PR against https://github.com/servo/media/pull/393 and see if that works?

delan (approved):

mrobinson:
Thanks for the review!

mrobinson:
> @mrobinson maybe you want to open a PR against #393 and see if that works?

I'm happy to give it a shot!

PR #407, opened 2023-11-16
Upgrade gstreamer dependency 2

ceyusa:
Rebase and update of https://github.com/servo/media/pull/393

Just to test CI

### DIFF
--- a/backends/gstreamer/lib.rs
+++ b/backends/gstreamer/lib.rs
@@ -15,7 +15,7 @@ pub mod webrtc;
 use device_monitor::GStreamerDeviceMonitor;
 use gst::prelude::*;
 use ipc_channel::ipc::IpcSender;
-use lazy_static::lazy_static;
+use glib::once_cell::sync::Lazy;
 use log::warn;
 use media_stream::GStreamerMediaStream;
 use mime::Mime;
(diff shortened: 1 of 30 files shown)


backends/gstreamer/lib.rs
```diff
@@ -103,7 +103,17 @@ impl GStreamerBackend {
             })
             .unwrap();
 
-        thread::spawn(|| glib::MainLoop::new(None, false).run());
+        // from GStreamer 1.19.1 to 1.22.7 glib's MainLoop is required
+        {
+            fn make_version(major: u32, minor: u32, micro: u32) -> u32 {
+                major << 29 | minor << 22 | micro << 12
```
sdroege, line 109: Where do these shifts come from? So odd numbers :D

backends/gstreamer/lib.rs
```diff
@@ -103,7 +103,17 @@ impl GStreamerBackend {
...
+            }
+            let (major, minor, micro, _) = gst::version();
+            let gst_version = make_version(major, minor, micro);
+            if  gst_version >= make_version(1, 19, 1) && gst_version <= make_version(1, 22, 7) {
```
sdroege, line 113: FWIW, you can also just use the `Ord` impl on tuples

backends/gstreamer/lib.rs
ceyusa: from vulkan headers  ;)

backends/gstreamer/lib.rs
ceyusa: rust is so lovely :)

rafaelcaricio (approved):

ceyusa:
closing as it's not required anymore...

PR #409, opened 2023-11-18
Remove unused legacy CI configuration

mrobinson:
These are no longer used after the switch to GitHub Actions.


atouchet:
Should Travis CI be removed as well?

mrobinson:
> Should Travis CI be removed as well?

Yep. I've gone ahead and added that change here.

jdm (approved):

PR #424, opened 2024-08-12
Handle errors from AudioRenderThread::start

Taym95:
Part of https://github.com/servo/servo/issues/32986, in order to handle errors in Servo, we need to remove panics and return errors.

### DIFF
--- a/audio/context.rs
+++ b/audio/context.rs
@@ -156,14 +156,13 @@ impl AudioContext {
             })
             .expect("Failed to spawn AudioRenderThread");
 
-        let thread_result = result_receiver.try_recv();
+        let thread_result = result_receiver
+            .recv()
+            .expect("Failed to receive result from AudioRenderThread");
 
-        let inner_result = match thread_result {
-            Ok(inner) => inner,
-            Err(_) => return Err(AudioSinkError::StateChangeFailed),
-        };
-
-        inner_result.map_err(|e| e.into())?;
+        if let Err(e) = thread_result {
+            return Err(e);
+        }
 
         Ok(Self {
             id,
--- a/audio/context.rs
+++ b/audio/context.rs
@@ -10,6 +10,8 @@ use std::sync::{Arc, Mutex};
 use std::thread::Builder;
 use AudioBackend;
 
+use crate::sink::AudioSinkError;
+
 /// Describes the state of the audio context on the control thread.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum ProcessingState {
@@ -130,7 +132,7 @@ impl AudioContext {
         client_context_id: &ClientContextId,
         backend_chan: Arc<Mutex<Sender<BackendMsg>>>,
         options: AudioContextOptions,
-    ) -> Self {
+    ) -> Result<Self, AudioSinkError> {
         let (sample_rate, channels) = match options {
             AudioContextOptions::RealTimeAudioContext(ref options) => (options.sample_rate, 2),
             AudioContextOptions::OfflineAudioContext(ref options) => {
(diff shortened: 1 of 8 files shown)


Taym95:
@jdm FYI

jdm:
Oops, I should have mentioned that there was someone else already looking into https://github.com/servo/servo/issues/32986.

jdm:
Since this PR exists, you should keep going with it. Thanks for looking into this!

audio/context.rs
```diff
@@ -143,13 +145,24 @@ impl AudioContext {
...
         let listener = graph.listener_id();
+
+        let (result_sender, result_receiver) = mpsc::channel();
         Builder::new()
             .name("AudioRenderThread".to_owned())
             .spawn(move || {
-                AudioRenderThread::start::<B>(receiver, sender_, sample_rate, graph, options);
+                let result =
+                    AudioRenderThread::start::<B>(receiver, sender_, sample_rate, graph, options);
+                result_sender.send(result).unwrap();
```
jdm, line 155: Let's ignore any error from sending here instead of using unwrap.

backends/dummy/lib.rs
```diff
@@ -94,14 +94,11 @@ impl Backend for DummyBackend {
...
+    ) -> Result<Arc<Mutex<AudioContext>>, AudioSinkError> {
         let (sender, _) = mpsc::channel();
         let sender = Arc::new(Mutex::new(sender));
-        Arc::new(Mutex::new(AudioContext::new::<Self>(
-            0,
-            &ClientContextId::build(1, 1),
-            sender,
-            options,
+        Ok(Arc::new(Mutex::new(
+            AudioContext::new::<Self>(0, &ClientContextId::build(1, 1), sender, options).unwrap(),
```
jdm, line 101: Can we use ? instead of unwrap here?

backends/gstreamer/lib.rs
```diff
@@ -184,18 +184,25 @@ impl Backend for GStreamerBackend {
...
-            id,
-            client_context_id,
-            self.backend_chan.clone(),
-            options,
-        )));
-        let mut instances = self.instances.lock().unwrap();
-        let entry = instances.entry(*client_context_id).or_insert(Vec::new());
-        entry.push((id, Arc::downgrade(&context).clone()));
-        context
+        let context_result =
```
jdm, line 189: Can we use ? here instead of this new match?

servo-media/lib.rs
```diff
@@ -6,6 +6,7 @@ pub extern crate servo_media_webrtc as webrtc;
 
 extern crate once_cell;
 
+use audio::sink::AudioSinkError;
```
jdm, line 9: Nit: move this with the other audio imports further down.

jdm (approved):

audio/context.rs
```diff
@@ -156,11 +156,14 @@ impl AudioContext {
             })
             .expect("Failed to spawn AudioRenderThread");
 
-        let thread_result: Result<(), AudioSinkError> = result_receiver.recv().unwrap();
+        let thread_result = result_receiver.try_recv();
```
jdm, line 159: This won't work quite right. The recv call is necessary because we need to block until we receive either a success or fail signal on the channel; if we're tiing out in tests, it means that we need to find a way to split the work of setting up the audio backend so that we can signal a success before proceeding with running it synchronously.

Taym95: okey got it, I will revert last commit and try to change how we do the setup.

Taym95: @jdm should we merge this one?

jdm: I don't think we can merge this without fixing the timeout issue you found, and I think the fix needs to be in this repository.

audio/render_thread.rs
```diff
@@ -160,10 +160,21 @@ impl AudioRenderThread {
...
-            .expect("Could not start audio render thread");
-        thread.event_loop(event_queue)
+        let mut thread =
+            match Self::prepare_thread::<B>(sender.clone(), sample_rate, graph, options) {
+                Ok(thread) => {
+                    let _ = init_sender.send(Ok(()));
+                    thread
+                }
+                Err(e) => {
+                    let _ = init_sender.send(Err(e));
```
Taym95, line 172: @jdm Thanks for the guidance! I refactored the code to send an error when prepare_thread returns an error. The tests are no longer timing out locally. Can you please review it again? Thanks!

Taym95: Hmm, it's not timing out locally, but it's still timing out on CI:

> Test results for linux-wpt-layout-2020 from try job ([#10431866463](https://github.com/Taym95/servo/actions/runs/10431866463)):
> 
> Flaky unexpected result (17)
> Stable unexpected results that are known to be intermittent (13)
> Stable unexpected results (1)
> * TIMEOUT [expected OK] `/fetch/metadata/audio-worklet.https.html`

Taym95: never-mind I was using old commit on CI!

jdm (approved):
Nicely done!

PR #427, opened 2024-10-17
Fix busted CI.

jdm:
This copies the result task from https://github.com/servo/ipc-channel/blob/main/.github/workflows/main.yml so we don't keep merging code that fails to build, then fixes the existing build errors.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -84,12 +84,12 @@ jobs:
   build_result:
     name: Result
     runs-on: ubuntu-latest
+    if: always()
     needs: ["Build"]
     steps:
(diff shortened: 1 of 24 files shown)


mukilan (approved):

sagudev:
CI is running for 4h but still no result.

Manishearth (approved):

jdm:
The latest commit should fix the CI timeout. I'm not sure how a shutdown deadlock been undetected for five years since https://github.com/servo/media/commit/8ccf3a4e18d3fa52215ba76f05550bed020596f2, but this could be a cause of pages going unresponsive in Servo too!

PR #429, opened 2025-01-16
Explicitly specify edition=2021 for all crates

simonwuelker:
Currently most crates don't specify a rust edition in their `Cargo.toml` files, meaning they are compiled using the 2015 edition. Cargo emits a warning when compiling them.

Minor changes to the imports in `audio` and `servo-media-derive` where necessary, for other crates its a simple change in `Cargo.toml`.

I also renamed `.cargo/config` to `.cargo/config.toml` (which is the "new" standard).

This fixes all cargo warnings emitted during a full build.


jdm (approved):
Thank you!

backends/gstreamer/lib.rs
```diff
@@ -73,6 +73,7 @@ impl GStreamerBackend {
         // See https://github.com/servo/media/pull/393.
         let needs_background_glib_main_loop = {
             let (major, minor, micro, _) = gst::version();
+            println!("glib version {major} {minor} {micro}");
```
mukilan, line 76: @simonwuelker, is this `println!`  here deliberate or should it be an `info!` or `debug!` invocation instead? I noticed "glibc version 1 24 7" being printed to stdout after updating my main branch and I was confused as it used to be silent before.

simonwuelker: Oops sorry, that is not intentional :grimacing:

simonwuelker: https://github.com/servo/media/pull/432

PR #430, opened 2025-01-17
Make ServoMedia::get infallible

simonwuelker:
It was already infallible before, but returned
a Result type for unknown reasons. The relevant
Code hasn't been touched for 7 years.

I think the original intention was to make backend *creation* fallible, but as mentioned before
it currently isn't. And even if it *were* fallible it seems more reasonable to return a Result from
ServoMedia::init instead.


jdm (approved):
Yeah, this is a bit silly!

simonwuelker:
Should we release a new crate version given that this is a breaking change or do we not care about that since `servo-media` isn't on crates.io?

jdm:
We don't care about that since it's never been published.

PR #431, opened 2025-01-19
Remove dependency on `lazy_static`

simonwuelker:
The gstreamer/webrtc crates were depending
    on lazy_static, but not using it.
    
In the case of the streams crate, lazy_static
can be replaced by the LazyLock struct from
    the standard library.
    (https://doc.rust-lang.org/std/sync/struct.LazyLock.html)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2545,7 +2545,6 @@ dependencies = [
  "gstreamer-video",
  "gstreamer-webrtc",
  "ipc-channel",
- "lazy_static",
  "log 0.4.17",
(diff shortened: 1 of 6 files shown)


nicoburns (approved):
Nice. There's a bunch of these across the Servo codebase! It's a shame it's a HashMap. If it was a type with a `const` constructor like a `BTreeMap` or a `Vec` then you could get rid of the `LazyLock` entirely.

simonwuelker:
Yeah, I was surprised seeing that `HashMap::new` is not const

PR #435, opened 2025-04-15
extend player api to support seekable

rayguo17:
extend player trait to have `fn seekable()`
if the source is `StreamType::Seekable` return its duration, else return `buffered` ranges.
related [spec](https://html.spec.whatwg.org/multipage/media.html#dom-media-seekable), 

Related Issue: https://github.com/servo/servo/issues/22297


jdm (approved):
Looks good!

PR #444, opened 2025-08-07
initial stub for ohos backend

rayguo17:
As a starting point for https://github.com/servo/media/issues/443. This PR add initial Backend Init logic that is mutual to all backend implementing `Backend` Trait. 

cc @jschwe @jdm @xiaochengh


rayguo17:
> Just note that this is not in workspace, hence build errors are hidden from CI.

Thanks for spotting, Let me try to fix it.

PR #457, opened 2025-11-01
gstreamer: Use weak ptr to renderer in sink callbacks

tharkum:
Passing weak pointers to the audio/video renderer in the audio/video sink callbacks to exclude circular references which will allow to release the high level renderer and it's associated resources (audio buffers and video frames).

Fixes (partially): https://github.com/servo/servo/issues/43387

### DIFF
--- a/backends/gstreamer/player.rs
+++ b/backends/gstreamer/player.rs
@@ -641,7 +641,7 @@ impl GStreamerPlayer {
                     if let Some(video_renderer) = weak_video_renderer.upgrade() {
                         video_renderer.lock().unwrap().render(frame);
                     } else {
-                        return Err(gst::FlowError::Error);
+                        return Err(gst::FlowError::Flushing);
                     };
 
                     let _ = notify!(observer, PlayerEvent::VideoFrameUpdated);


backends/gstreamer/player.rs
```diff
@@ -621,13 +629,21 @@ impl GStreamerPlayer {
...
                         .map_err(|_| gst::FlowError::Error)?;
-                    video_renderer.lock().unwrap().render(frame);
+
+                    if let Some(video_renderer) = weak_video_renderer.upgrade() {
+                        video_renderer.lock().unwrap().render(frame);
+                    } else {
+                        return Err(gst::FlowError::Flushing);
+                    };
+
                     let _ = notify!(observer, PlayerEvent::VideoFrameUpdated);
```
sdroege: Should this still happen even if `video_renderer` is gone?

sdroege (approved):
Apart from that looks correct to me

backends/gstreamer/player.rs
```diff
@@ -462,7 +462,9 @@ impl GStreamerPlayer {
             pipeline.set_property("audio-sink", &audio_sink);
 
             let audio_sink = audio_sink.dynamic_cast::<gst_app::AppSink>().unwrap();
-            let audio_renderer_ = audio_renderer.clone();
+
+            let weak_audio_renderer = Arc::downgrade(&audio_renderer);
```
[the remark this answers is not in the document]
tharkum: I tried it for video_renderer (same for audio_renderer) but have compilation error

```
error[E0277]: the size for values of type `(dyn servo_media_player::video::VideoFrameRenderer + 'static)` cannot be known at compilation time
   --> /home/workspace/projects/servo/media/backends/gstreamer/player.rs:629:34
    |
629 |               let render_sample = {glib::clone!(
    |  __________________________________^
630 | |                 #[strong(rename_to = observer)]
631 | |                 self.observer,
632 | |                 #[strong(rename_to = render)]
...   |
654 | |             )};
    | |_____________^ doesn't have a size known at compile-time
    |
    = help: within `std::sync::Mutex<(dyn servo_media_player::video::VideoFrameRenderer + 'static)>`, the trait `std::marker::Sized` is not implemented for `(dyn servo_media_player::video::VideoFrameRenderer + 'static)`
    = help: the trait `glib::clone::Downgrade` is implemented for `std::sync::Arc<T>`
note: required because it appears within the type `std::sync::Mutex<(dyn servo_media_player::video::VideoFrameRenderer + 'static)>`
   --> /home/v00863305/.rustup/toolchains/1.91.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sync/poison/mutex.rs:227:12
    |
227 | pub struct Mutex<T: ?Sized> {
    |            ^^^^^
    = note: required for `std::sync::Arc<std::sync::Mutex<(dyn servo_media_player::video::VideoFrameRenderer + 'static)>>` to implement `glib::clone::Downgrade`
    = note: this error originates in the macro `glib::clone` (in Nightly builds, run with -Z macro-backtrace for more info)
```

backends/gstreamer/player.rs
tharkum: Yes, no point to hold strong ref to video renderer too long, because anyway during marshalling this event the media player could be drop and message will be discarded on half-way

PR #459, opened 2025-11-05
Apply servo style rust formatting

tharkum:
Added the rust formatting configuration options similar to servo/servo and servo/stylo with the format style edition is inferred from the individual package edition (2021).


tharkum:
@jdm < Please take a look.

PR #470, opened 2025-11-22
Build gstreamer in CI

arihant2math:
Needed for gstreamer 1.26 support.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,49 +17,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04]
+        os: [ubuntu-latest]
         rust: [stable]
         target: ["default"]
     steps:
     - uses: actions/checkout@v2
-    - name: Install deps on linux
-      if: startsWith(matrix.os, 'ubuntu')
-      run: |
-        sudo bash -c 'apt-add-repository -y https://mirrors.kernel.org/ubuntu'
-        sudo apt update
-        sudo apt install -y \
-          gstreamer1.0-libav \
-          gstreamer1.0-plugins-bad \
-          gstreamer1.0-plugins-base \
-          gstreamer1.0-plugins-good \
-          gstreamer1.0-plugins-ugly \
-          gstreamer1.0-tools \
-          libasound2-plugins \
-          libfaad2 \
-          libffi7 \
-          libfftw3-single3 \
-          libges-1.0-dev \
-          libgstreamer-plugins-bad1.0-dev\
-          libgstreamer-plugins-base1.0-dev \
-          libgstreamer1.0-dev \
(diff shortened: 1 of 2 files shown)


etc/build_gstreamer.sh
```diff
@@ -0,0 +1,22 @@
...
+if [ -z "$1" ]; then
+    echo "Usage: $0 <path>"
+    exit 1
+fi
+
+# Change directory to the provided path
+cd "$1" || exit 1
+
+# Clone gstreamer repository
+git clone --depth=1 https://gitlab.freedesktop.org/gstreamer/gstreamer.git
```
nirbheek:
```suggestion
git clone --depth=1 -b 1.26.8 https://gitlab.freedesktop.org/gstreamer/gstreamer.git
```

etc/build_gstreamer.sh
```diff
@@ -0,0 +1,22 @@
...
+cd gstreamer || exit 1
+
+# Build gstreamer
+meson setup ../gst-build
```
nirbheek:
```suggestion
meson setup --wrap-mode=nodownload ../gst-build
```

This will ensure that meson doesn't download any dependencies using wrap files. You should continue to use ubuntu for dependencies.

Also, once you have gotten the build going, if you want to optimize the build time, you can use `-Dauto_features=disabled` and then manually enable the features/plugins you need. Note that this can be tricky.

.github/workflows/rust.yml
```diff
@@ -17,49 +17,17 @@ jobs:
...
-          libtdb1 \
-          libtheora0 \
-          libtwolame0 \
-          libwayland-egl1-mesa \
-          libwebrtc-audio-processing1 \
-          pulseaudio \
-          pulseaudio-utils \
-          webp
+    - name: Install dependencies
+      run: sudo apt-get update && sudo apt-get install -y libasound2-dev libjack-dev libpulse-dev ninja-build gcc
```
nirbheek: I would recommend keeping around the various plugin deps, with development packages: libfaad-dev libmpg123-dev etc. That will keep the build time low, and also ensure that you build the plugins you expect. Not everything has wrap files, for example wayland and webp.

Just make sure you don't accidentally pull in gstreamer itself using one of the deps. I would place an assert after this checking that `libgstreamer-1.0.so.*` is not in libdir.

nirbheek:
```
2025-11-22T19:33:50.8169861Z PermissionError: [Errno 13] Permission denied: '/usr/local/share/gstreamer-1.0'
2025-11-22T19:33:50.8172027Z Installing subdir /home/runner/work/media/media/gstreamer/gstreamer/subprojects/gst-devtools/validate/data/scenarios/includes to /usr/local/share/gstreamer-1.0/validate/scenarios/includes
```

You can just disable gst-devtools with `-Ddevtools=disabled`.

PR #471, opened 2025-11-22
chore: switch to parking_lot Mutex

webbeef:
This is needed to move Servo to `parking_lot` since some `Mutex` leaks through public apis.


jdm (approved):

webbeef:
@jdm looks like someone needs to push some button to merge :)

PR #476, opened 2025-12-05
Fix `bufferSource.loop = true` on `extreme edge case`

Taym95:
Fixes: https://github.com/servo/servo/issues/41073

Previously: 
the moment we set bufferSource.loop = true, the node hit the early “extreme edge case” check, decided “I refuse to output”, triggered onended, and returned silence. So the loop never got a chance to run.

Now:

- Early bailout is removed(no idea why was there in first place).
- The per tick offset is guaranteed to be a sensible value with respect to the loop range (wrapped if extreme).
- The normal mixing code runs, updating pos each tick and respecting the loop region.

### DIFF
--- a/audio/buffer_source_node.rs
+++ b/audio/buffer_source_node.rs
@@ -303,15 +303,6 @@ impl AudioNodeEngine for AudioBufferSourceNode {
 
                 for sample in data {
                     if duration <= 0. {
-                        println!(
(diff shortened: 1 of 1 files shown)


audio/buffer_source_node.rs
```diff
@@ -292,6 +303,15 @@ impl AudioNodeEngine for AudioBufferSourceNode {
 
                 for sample in data {
                     if duration <= 0. {
+                        println!(
```
jdm: Let's use `log::warning!` instead.

PR #477, opened 2025-12-05
Fix IIRFilter crash on silent blocks

Taym95:
fixes https://github.com/servo/servo/issues/41085


jdm (changes requested):
I was confused why these changes worked and the original code did not. I have come to the conclusion that Chunk::explicit_silence contains a bug that this PR is working around, and I would rather see us fix the bug properly.

https://github.com/servo/media/blob/b0d3b744b0114fb66a05dcb80c209c0f7b94ac8e/audio/block.rs#L45 is written with an assumption that `blocks` contains one element, but we never actually push that element. The crash should go away if we fix that.

PR #478, opened 2025-12-17
Add feature flags to for X11 and Wayland dependencies on Linux

stshine:
Add 'glx' and 'wayland' feature flags so that we can choose to disable the features if they are not enabled on gstreamer.


stshine:
Hi :-)
Commit updated.

Issue #480, opened 2025-12-18
`GStreamerPlayer` resumes playback on `resume` even if it wasn't playing on `suspend`


yerke:
This issue is the root cause of https://github.com/servo/servo/issues/41281.

Currently `GStreamerPlayer` resumes playback on `resume` even if it wasn't playing on `suspend`.

I have a PR to fix it.

PR #481, opened 2025-12-18
GStreamerPlayer should resume playback only if it was playing on suspend

yerke:
GStreamerPlayer should resume playback only if it was playing on suspend

Fixes https://github.com/servo/media/issues/480 and https://github.com/servo/servo/issues/41281.

### Test case 1: does not resume if it was not playing on suspend

```
<!doctype html>
<audio preload="auto" networkstate="0"><source src="https://fairysvoice.net/games/Fire_Triss/html5game/an1.mp3" type="audio/mp3"></audio>
<a href="https://example.org">CLICK THIS THEN GO BACK</a>
```

Steps:
- Music should not be playing when the page loads.
- Click the link, then hit the back button.
- Music should not be playing.

### Test case 2: resumes if it was playing on suspend

```
<!doctype html>
<audio preload="auto" autoplay="true"><source src="https://fairysvoice.net/games/Fire_Triss/html5game/an1.mp3" type="audio/mp3"></audio>
<a href="https://example.org">CLICK THIS THEN GO BACK</a>
```

Steps:
- Music should be playing when the page loads.
- Click the link, then hit the back button.
- Music should resume playing.

### DIFF
--- a/backends/dummy/lib.rs
+++ b/backends/dummy/lib.rs
@@ -158,6 +158,9 @@ impl Player for DummyPlayer {
     fn paused(&self) -> bool {
         true
     }
+    fn can_resume(&self) -> bool {
+        true
+    }
 
     fn stop(&self) -> Result<(), PlayerError> {
         Ok(())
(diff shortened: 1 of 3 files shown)


jdm (approved):
Nice!

